PR #45, opened 2023-06-26
Updated nodes package

evgenLevin:


pkg/nodes/node.go
```diff
@@ -78,12 +82,38 @@ func (builder *NodeBuilder) Exists() bool {
...
+// Get returns node object if found.
+func (builder *NodeBuilder) Get() (*v1.Node, error) {
+	if valid, err := builder.validate(); !valid {
+		return nil, err
+	}
+
+	glog.V(100).Infof("Collecting node object %s", builder.Definition.Name)
+
+	node := &v1.Node{}
+	err := builder.apiClient.Get(context.TODO(), goclient.ObjectKey{
```
trewest, line 103: shouldn't you use the typed client interface rather than the generic interface? Like was done in the Exists method previously?

pkg/nodes/node.go
```diff
@@ -189,6 +219,32 @@ func (builder *NodeBuilder) ExternalIPv4Network() (string, error) {
 	return extNetwork.IPv4, nil
 }
 
+// WaitUntilNodeStable waits for the duration of the defined timeout or until the
+// node gets to a stable - not unschedulable.
+func (builder *NodeBuilder) WaitUntilNodeStable(timeout time.Duration) error {
+	glog.V(100).Infof("Waiting for the defined period until node %s is not unschedulable",
```
trewest, line 225: please add a validate block at the top of this method

pkg/nodes/node.go
```diff
@@ -189,6 +219,32 @@ func (builder *NodeBuilder) ExternalIPv4Network() (string, error) {
...
+		builder.Definition.Name)
+
+	if builder.errorMsg != "" {
```
trewest, line 228: this will be redundant when the validate block is added

pkg/nodes/node.go
```diff
@@ -189,6 +219,32 @@ func (builder *NodeBuilder) ExternalIPv4Network() (string, error) {
 	return extNetwork.IPv4, nil
 }
 
+// WaitUntilNodeStable waits for the duration of the defined timeout or until the
+// node gets to a stable - not unschedulable.
+func (builder *NodeBuilder) WaitUntilNodeStable(timeout time.Duration) error {
```
trewest, line 224: I think a more appropriate name for this method is `WaitUntilSchedulable`. node is the name of the pkg so no need to add it to the method name.

trewest: You can then update subsequent logs/comments from `not unschedulable` to just `schedulable`

trewest (changes requested):
Few comments

pkg/nodes/node.go
```diff
@@ -189,6 +219,32 @@ func (builder *NodeBuilder) ExternalIPv4Network() (string, error) {
...
+
+	// Polls every retryInterval to determine if the node is not unschedulable.
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		node, err := builder.Get()
+
+		if err != nil {
+			return false, nil
+		}
+
+		if !node.Spec.Unschedulable {
```
kononovn, line 240: I think you need to make sure that node is in Schedulable state for some specific period of time, sometimes we mark node as schedulable and them after 5 seconds move to unscheduled

kononovn: Please refer to: https://github.com/openshift-kni/eco-goinfra/blob/b2486f52938008d0fa468e4e38a852ec60e9027f/pkg/mco/mcp.go#L248

kononovn (changes requested):
Please fix few comments above.

kononovn:
Commit message is very generic. Can you please add more specific info

evgenLevin:
@trewest @kononovn  PTAL one more time. Thanks!

pkg/nodes/node.go
kononovn: No you need to add 1 more PollImmediate function with additional stable timeout. Please check the reference link that I've share with you

kononovn (changes requested):
Please check my comment above

PR #47, opened 2023-06-26
Added WithInterfaceAndNamespace func to pod pkg

kononovn:


pkg/pod/network.go
```diff
@@ -13,3 +13,14 @@ func StaticIPAnnotation(name string, ipAddr []string) []*multus.NetworkSelection
 		},
 	}
 }
+
+// StaticIPAnnotationWithInterfaceAndNamespace defines static ip address, interface name and namespace
+// network annotation for pod object.
+func StaticIPAnnotationWithInterfaceAndNamespace(
+	name, namespace, intName string, ipAddr []string) []*multus.NetworkSelectionElement {
```
trewest, line 20: can you please define each parameter on a separate line

pkg/pod/network.go
```diff
@@ -13,3 +13,14 @@ func StaticIPAnnotation(name string, ipAddr []string) []*multus.NetworkSelection
...
+	baseAnnotation := StaticIPAnnotation(name, ipAddr)
```
trewest, line 21: Should you check name, namespace, and intName are not "" ?

trewest (commented):
Couple comments

pkg/pod/network.go
kononovn: According to convention from README it's OK

```
func Function(
    argInt1, argInt2 int, argString1, argString2 string, argSlice1, argSlice2 []string) {
	
}
```

pkg/pod/network.go
kononovn: Yes right. We need to add validation function here. I think it'll be my next PR because this change is going to affect both projects it would require multiple PRs in to diff project.

kononovn:
@trewest PTAL

trewest (approved):
/lgtm

PR #48, opened 2023-06-26
Added tap plugin and linkInContainer to nad pkg

kononovn:


pkg/nad/masterplugin.go
```diff
@@ -101,6 +101,20 @@ func (plugin *MasterMacVlanPlugin) WithIPAM(ipam *IPAM) *MasterMacVlanPlugin {
 	return plugin
 }
 
+// WithLinkInContainer defines MasterMacVlan plugin using linkInContainer feature.
+func (plugin *MasterMacVlanPlugin) WithLinkInContainer() *MasterMacVlanPlugin {
+	glog.V(100).Infof("Adding linkInContainer configuration to MasterMacVlanPlugin")
+
+	if plugin.masterPlugin == nil {
```
trewest, line 108: Maybe safe to just also check if plugin == nil also

trewest (commented):
Small comment

pkg/nad/masterplugin.go
kononovn: According to function NewMasterMacVlanPlugin this condition (plugin == nil) will be always false

kononovn:
@trewest PTAL

trewest (approved):
/lgtm

PR #50, opened 2023-06-26
Added NMState struct

evgenLevin:


pkg/nmstate/nmstate.go
```diff
@@ -1 +1,223 @@
...
+// the cluster and the NMState definitions.
+type Builder struct {
+	Definition *nmstateV1.NMState
+	Object     *nmstateV1.NMState
+	apiClient  *clients.Settings
+	errorMsg   string
+}
+
+// NewBuilder creates a new instance of Builder.
+func NewBuilder(apiClient *clients.Settings, name string) *Builder {
```
kononovn, line 29: Do we have scenario when we need to create this object manually? I think we can remove this function.

pkg/nmstate/nmstate.go
```diff
@@ -1 +1,223 @@
...
+	return builder, err
+}
+
+// Delete removes NMState object from a cluster.
+func (builder *Builder) Delete() (*Builder, error) {
+	if valid, err := builder.validate(); !valid {
+		return builder, err
+	}
+
+	glog.V(100).Infof("Deleting the NMState object %s",
```
kononovn, line 113: No need to add extra line. Change this glog to oneline

pkg/nmstate/nmstate.go
```diff
@@ -1 +1,223 @@
...
+		builder.Definition.Name,
+	)
+
+	if !builder.Exists() {
```
kononovn, line 117: Please remove this check. We basically don't care if NS exist in case if we want to remove it. If it's important you can leave the log message but no need to return error.

kononovn: Just add something like:

	if !builder.Exists() {
		return nil
	}

kononovn (changes requested):
looks good just few minor changes

pkg/nmstate/nmstate.go
evgenLevin: removed.

evgenLevin:
@kononovn PTAL

pkg/nmstate/nmstate.go
```diff
@@ -1 +1,221 @@
...
+	Definition *nmstateV1.NMState
+	// Created NMState object on the cluster.
+	Object *nmstateV1.NMState
+	// API client to interact with the cluster.
+	apiClient *clients.Settings
+	// errorMsg is processed before NMState object is created.
+	errorMsg string
+}
+
+// NewBuilder creates a new instance of Builder.
```
kononovn, line 32: Please update to 
// NewBuilder creates a new instance of nmstate Builder.

pkg/nmstate/nmstate.go
```diff
@@ -1 +1,221 @@
...
+			return builder.Create()
+		}
+	}
+
+	return builder, err
+}
+
+// PullNMstate retrieves an existing NMState object from the cluster.
+func PullNMstate(apiClient *clients.Settings, name string) (*Builder, error) {
+	glog.V(100).Infof("Pulling NMState object name:%s", name)
```
kononovn, line 165: Please add space 
glog.V(100).Infof("Pulling NMState object name: %s", name)

kononovn (changes requested):
Please fix 2 minor comments

kononovn (approved):
lgtm

pkg/nmstate/nmstate.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	if name == "" {
+		glog.V(100).Infof("The name of the NMState is empty")
+
+		builder.errorMsg = "NMState 'name' cannot be empty"
+	}
+
+	if !builder.Exists() {
+		return nil, fmt.Errorf("NMState oject %s doesn't exist", name)
```
trewest, line 183: nit: object

trewest (dismissed):
Just one small nit, other than that lgtm

evgenLevin:
@trewest @kononovn PTAL

kononovn (approved):

PR #54, opened 2023-06-29
add vendor for PAO pkg folder

gkopels:


pkg/pao/pao.go
```diff
@@ -0,0 +1,70 @@
+package pao
```
trewest, line 1: Did you mean to leave this file in? If so, can we make updates so that it adheres to the standard builder definitions and factories that we have in other pkgs? Please reference other pkgs for examples

trewest (changes requested):
Please address comment

kononovn:
This PR is covered in https://github.com/openshift-kni/eco-goinfra/pull/58 . I close this one

PR #58, opened 2023-07-05
Added nto package with PerformanceProfile resource

kononovn:


pkg/nto/performanceprofile.go
```diff
@@ -1 +1,351 @@
...
+func (builder *Builder) Create() (*Builder, error) {
+	if valid, err := builder.validate(); !valid {
+		return builder, err
+	}
+
+	glog.V(100).Infof("Creating PerformanceProfile %s ", builder.Definition.Name)
+
+	var err error
+	if !builder.Exists() {
+		err = builder.apiClient.Create(context.TODO(), builder.Definition)
```
evgenLevin, line 257: I think we should define `builder.Object` here

evgenLevin (commented):
just one comment

gkopels (commented):
lgtm

PR #61, opened 2023-07-06
nad pkg added Update func to Builder

kononovn:


pkg/nad/builder.go
```diff
@@ -155,6 +155,26 @@ func (builder *Builder) Delete() error {
...
+		return builder, err
+	}
+
+	glog.V(100).Infof("Updating NetworkAttachmentDefinition %s in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	var err error
+
+	builder.Definition.CreationTimestamp = metaV1.Time{}
+	builder.Definition.ResourceVersion = builder.Object.ResourceVersion
```
evgenLevin, line 170: Why do we want to keep resorceversion?

kononovn: Yes, I'm not sure but it looks like it's required

evgenLevin: We don't do that  in the node package
https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/nodes/node.go#L63

PR #65, opened 2023-07-10
Add cpu and memory limit to Pod builder

gkopels:


pkg/clients/clients.go
```diff
@@ -212,6 +217,14 @@ func SetScheme(crScheme *runtime.Scheme) error {
 		return err
 	}
 
+	if err := nmstatev1.AddToScheme(crScheme); err != nil {
```
kononovn, line 220: According to commit message it's not related to your PR

pkg/pod/pod.go
```diff
@@ -487,6 +489,32 @@ func (builder *Builder) WithTolerationToMaster() *Builder {
...
+	builder.isMutationAllowed("resource spec")
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	for index := range builder.Definition.Spec.Containers {
+		builder.Definition.Spec.Containers[index].Resources = v1.ResourceRequirements{
+			Limits: v1.ResourceList{
+				v1.ResourceCPU:    resource.MustParse(strconv.Itoa(2)),
```
kononovn, line 509: It shouldn't be hard-coded

pkg/clients/clients.go
gkopels: I didnt make any changes to this file. I believe I incorrectly created a commit including the previous PR. Need to remove.

kononovn:
closing PR because relevant tests are going to be added upstream.

PR #76, opened 2023-07-26
Fix KMM WithImageRepoSecret function

cdvultur:
In the current format, the call to WithImageRepoSecret fails with:

```
  [PANICKED] Test Panicked
  In [It] at: /home/cvultur/sdk/go1.19.3/src/runtime/panic.go:260 @ 07/26/23 11:59:45.932

  runtime error: invalid memory address or nil pointer dereference
```

Verified the solution in the test itself and seems to work properly


trewest (approved):
/lgtm

wabouhamad (approved):

wabouhamad (commented):
/lgtm

PR #79, opened 2023-07-27
Add copy method to pod pkg

trewest:
Adds a copy method to the pod pkg which can copy files from one of the pod's containers. Has two options:
1. With tar - creates a tar archive of a path inside the container (good for copying directories if a container supports it)
2. Without tar - copies the contents of a single file

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/pod/pod.go
```diff
@@ -401,6 +404,96 @@ func (builder *Builder) ExecCommand(command []string, containerName ...string) (
...
+		Proxier:    proxy,
+		PingPeriod: 0,
+	})
+
+	wrapper, err := rest.HTTPWrappersForConfig(builder.apiClient.Config, upgradeRoundTripper)
+	if err != nil {
+		return bytes.Buffer{}, err
+	}
+
+	if err != nil {
```
kononovn, line 473: Same if condition as above, Please remove it

trewest: Fixed

kononovn (approved):
lgtm

PR #84, opened 2023-08-02
pod pkg, fixing Exists function comment

vkolodny:
The initial comment seems to be copy pasted from namespace package

// Exists checks whether the given namespace exists.

Should be

//Exists checks whether the given pod exists.


kononovn (approved):
lgtm

kononovn (approved):
lgtm

PR #87, opened 2023-08-10
Fix for KMM's deviceplugin

cdvultur:
Fixing the : `  runtime error: invalid memory address or nil pointer dereference` while using `WithDevicePlugin*` functions

### DIFF
--- a/pkg/kmm/module.go
+++ b/pkg/kmm/module.go
@@ -149,6 +149,10 @@ func (builder *ModuleBuilder) WithDevicePluginVolume(name string, configMapName 
 		return builder
 	}
 
+	if builder.Definition.Spec.DevicePlugin == nil {
+		builder.Definition.Spec.DevicePlugin = &moduleV1Beta1.DevicePluginSpec{}
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/kmm/module.go
```diff
@@ -359,6 +367,10 @@ func (builder *ModuleBuilder) withServiceAccount(srvAccountName string, accountT
 	case "module":
 		builder.Definition.Spec.ModuleLoader.ServiceAccountName = srvAccountName
 	case "device":
```
qbarrand, line 369: I would probably define a new type for this:
```go
type ServiceAccountType string

const (
    ServiceAccountTypeModule ServiceAccountType = "module"
    ServiceAccountTypeDevice ServiceAccountType = "device"
)
```
The `switch` then becomes:
```go
	switch accountType {
	case ServiceAccountTypeModule:
		builder.Definition.Spec.ModuleLoader.ServiceAccountName = srvAccountName
	case ServiceAccountTypeDevice:
```

trewest (approved):
/lgtm

pkg/kmm/module.go
cdvultur: I think I will leave it as is for the moment, given that this will change the signature of the function, causing other refactorings

kononovn (approved):
lgtm

PR #90, opened 2023-08-21
extend client's interfaces for argocd apps

shaior:


shaior:
@kononovn Hi, it looks like the linter does not exclude the error shown from the linter - 
`issues:
  exclude:
    - "Can't run linter goanalysis_metalinter: inspect: failed to load package"`

PR #92, opened 2023-08-23
Added argocd package

shaior:


shaior:
I will create a new PR for that after client updates.

PR #102, opened 2023-09-01
rename bmh builder and file name

elenagerman:
rename bmh Builder to be unique
rename filename (this one includes methods related to the single object handling; at least one more file will be added - handle list of baremetalhosts)

### DIFF
--- a/pkg/bmh/baremetalhost.go
+++ b/pkg/bmh/baremetalhost.go
@@ -19,9 +19,9 @@ import (
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// Builder provides struct for the bmh object containing connection to
+// BmhBuilder provides struct for the bmh object containing connection to
 // the cluster and the bmh definitions.
-type Builder struct {
+type BmhBuilder struct {
 	Definition *bmhv1alpha1.BareMetalHost
 	Object     *bmhv1alpha1.BareMetalHost
 	apiClient  *clients.Settings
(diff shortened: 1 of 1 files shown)


kononovn:
@elenagerman This change will break most of test cases in test projects. Can we avoid this change ?

kononovn (changes requested):
@elenagerman This change will break most of test cases in test projects because we are changing public object that is in use in most of test suites. Can we avoid this change ?

elenagerman:
> @elenagerman This change will break most of test cases in test projects because we are changing public object that is in use in most of test suites. Can we avoid this change ?

@kononovn I didn't find any test yet which uses this library.
@trewest , for example, for some reason, uses bmhv1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1" instead of package:
https://github.com/openshift-kni/eco-gotests/blob/main/tests/assisted/ztp/operator/internal/tsparams/operatorvars.go
https://github.com/openshift-kni/eco-gotests/blob/main/tests/assisted/ztp/spoke/internal/tsparams/spokevars.go

I think it will be better to do it now then wait until it will be used by eco-gotests

kononovn:
> @kononovn Please let me know, if you prefer do not rename it anyway, we will close this MR and I will call my new method "list.go' instead of 'baremetalhosts.go'

You are right. This package is not in use so no problem with re-naming. I'll merge it

kononovn (approved):
lgtm

trewest (approved):
/lgtm

PR #101, opened 2023-08-31
Added clusteroperators package

elenagerman:

PR #103, opened 2023-09-03
nodes pkg: new methods added

elenagerman:
- IsReady: verify node condition Ready is True
- WaitToBeAllReadyFor: wait until all required nodes will reach the Ready state


elenagerman:
@kononovn @trewest @achuzhoy PTAL

pkg/nodes/node.go
```diff
@@ -189,6 +195,49 @@ func (builder *NodeBuilder) ExternalIPv4Network() (string, error) {
...
+// IsReady check if the Node is Ready.
+func (builder *NodeBuilder) IsReady() bool {
+	if valid, _ := builder.validate(); !valid {
+		return false
+	}
+
+	glog.V(100).Infof("Verify %s node availability", builder.Definition.Name)
+
+	var err error
+	builder.Object, err = builder.Pull()
```
kononovn, line 207: Please re-place this method by func Exists. It'll update builder.Object automatically

pkg/nodes/node.go
```diff
@@ -14,6 +15,11 @@ import (
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
```
kononovn, line 18: Please move this consts to ./pkg/nodes/nodes.go

kononovn (changes requested):
Looks good, Just 2 small comments to fix

pkg/nodes/node.go
elenagerman: done

pkg/nodes/node.go
elenagerman: done

usage/namespace/namespace.go
```diff
@@ -33,11 +33,11 @@ func deleteExistingExample() {
 	if apiClients == nil {
 		panic("Failed to load api client")
 	}
-	// Pull existing namespace from cluster
+	// Get existing namespace from cluster
```
kononovn, line 36: I thinks changed in [usage/namespace/namespace.go](https://github.com/openshift-kni/eco-goinfra/pull/103/files#diff-58fc6fcda40f9069be2b457da77e1ab6e146c5110a23fd5b3113a6331b735c93) were added by mistake in this PR

pkg/nodes/node.go
```diff
@@ -189,6 +188,46 @@ func (builder *NodeBuilder) ExternalIPv4Network() (string, error) {
...
+		if condition.Type == "Ready" {
+			return condition.Status == isTrue
+		}
+	}
+
+	return false
+}
+
+// Get fetches the defined Node from the cluster.
+func (builder *NodeBuilder) Get() (*v1.Node, error) {
```
kononovn, line 213: I think you can safely remove this function from commit. PullNode or Exists safely pull latest node struct in to builder.Object that you can later use in test cases.

usage/namespace/namespace.go
elenagerman: uff, here as well, sorry,it's not exactly me, for some reason it was changed when I did name refactoring for nodes

kononovn (approved):
lgtm

pkg/nodes/nodes.go
```diff
@@ -97,6 +105,50 @@ func (builder *Builder) ExternalIPv4Networks() ([]string, error) {
 	return ipV4ExternalAddresses, nil
 }
 
+// WaitToBeAllReadyFor waits for all nodes to be Ready for a time duration up to the timeout.
+func (builder *Builder) WaitToBeAllReadyFor(timeout time.Duration) (bool, error) {
```
trewest, line 109: Same comment as https://github.com/openshift-kni/eco-goinfra/pull/101/files#r1317940396, sorry if I wasn't clear

trewest (changes requested):
Just one comment

pkg/nodes/nodes.go
elenagerman: I answer on 101 thread, I don't think we need additional wait in this case. this will increase the execution time of this method

pkg/nodes/list.go
```diff
@@ -60,3 +68,98 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
 
 	return ipV4ExternalAddresses, nil
 }
+
+// WaitForAllNodesAreReady waits for all nodes to be Ready for a time duration up to the timeout.
+func WaitForAllNodesAreReady(apiClient *clients.Settings,
+	options v1.ListOptions,
```
trewest, line 74: Can you move options to the last parameter? We are going to change this to an optional param in a future PR and don't want to break any tests that use it

elenagerman: sure

pkg/nodes/node.go
```diff
@@ -189,6 +196,71 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+
+	glog.V(100).Infof("Verify %s node availability", builder.Definition.Name)
+
+	if !builder.Exists() {
+		return false, fmt.Errorf("%s node object doesn't exist", builder.Definition.Name)
+	}
+
+	for _, condition := range builder.Object.Status.Conditions {
+		if condition.Type == "Ready" {
+			if condition.Status == isTrue {
```
trewest, line 213: I think you should change this to:
```suggestion
			return condition.Status == isTrue, nil
```
err should be reserved for if an actual error occurs

pkg/nodes/node.go
```diff
@@ -189,6 +196,71 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+				return true, nil
+			}
+
+			return false,
+				fmt.Errorf("%s node not in Ready state; current status is: %s",
+					builder.Definition.Name, condition)
+		}
+	}
+
+	return false, fmt.Errorf("%s node is not in Ready state", builder.Definition.Name)
```
trewest, line 223: This error should say that the Ready condition could not be found

pkg/nodes/node.go
```diff
@@ -189,6 +196,71 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+}
+
+// WaitUntilInStatus waits for timeout duration or until node gets to a specific status.
+func (builder *Builder) WaitUntilInStatus(
+	status v1.NodeConditionType, timeout time.Duration) error {
```
trewest, line 228:
```suggestion
	conditionType v1.NodeConditionType, timeout time.Duration) error {
```

pkg/nodes/node.go
```diff
@@ -189,6 +196,71 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+				fmt.Errorf("%s node not in Ready state; current status is: %s",
+					builder.Definition.Name, condition)
+		}
+	}
+
+	return false, fmt.Errorf("%s node is not in Ready state", builder.Definition.Name)
+}
+
+// WaitUntilInStatus waits for timeout duration or until node gets to a specific status.
+func (builder *Builder) WaitUntilInStatus(
```
trewest, line 227:
```suggestion
func (builder *Builder) WaitUntilConditionTrue(
```

pkg/nodes/node.go
```diff
@@ -189,6 +196,71 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+	status v1.NodeConditionType, timeout time.Duration) error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	err := wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		if !builder.Exists() {
+			return false, fmt.Errorf("node object doesn't exist")
```
trewest, line 235: Can you add the node name to the error message?

pkg/nodes/list.go
```diff
@@ -60,3 +68,98 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
 
 	return ipV4ExternalAddresses, nil
 }
+
+// WaitForAllNodesAreReady waits for all nodes to be Ready for a time duration up to the timeout.
+func WaitForAllNodesAreReady(apiClient *clients.Settings,
+	timeout time.Duration,
+	options v1.ListOptions) (bool, error) {
+	glog.V(100).Infof("Waiting for all nodes in the Ready state up to duration of %v",
```
trewest, line 76:
```suggestion
	glog.V(100).Infof("Waiting for all nodes to be in the Ready state for up to a duration of %v",
```

pkg/nodes/list.go
```diff
@@ -60,3 +68,98 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+	// Here err is "timed out waiting for the condition"
+	glog.V(100).Infof("Not all nodes were found in the Ready State during availableDuration: %v",
+		err)
+
+	return false, err
+}
+
+// WaitForAllNodesToReboot waits for all nodes to start and finish reboot up to the timeout.
+func WaitForAllNodesToReboot(apiClient *clients.Settings,
+	options v1.ListOptions,
```
trewest, line 120: Please move options to the last argument in the function

pkg/nodes/list.go
```diff
@@ -60,3 +68,98 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+	timeout time.Duration) (bool, error) {
+	glog.V(100).Infof("Waiting for all nodes the list to reboot and return to the Ready condition")
```
trewest, line 122:
```suggestion
	glog.V(100).Infof("Waiting for all nodes in the list to reboot and return to the Ready condition")
```

pkg/nodes/list.go
```diff
@@ -60,3 +68,98 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+
+	nodesList, err := List(apiClient, options)
+	if err != nil {
+		glog.V(100).Infof("Failed to list all nodes due to %s", err.Error())
+
+		return false, err
+	}
+
+	readyNodes := []string{}
+	err = wait.PollImmediate(backoff, timeout, func() (done bool, err error) {
```
trewest, line 132: Does this work as expected? Just wondering how safe this function is as it seems like a race condition could be encountered easily

pkg/nodes/list.go
```diff
@@ -60,3 +68,98 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+		for _, node := range nodesList {
+			if !slices.Contains(readyNodes, node.Object.Name) {
+				err = node.WaitUntilNotReady(retryInterval)
+				if err != nil {
+					glog.V(100).Infof("%s node reboot not started yet", node.Object.Name)
+
+					return false, err
+				}
+
+				err = node.WaitUntilReady(timeout / 2)
```
trewest, line 142: Why split the timeout?

trewest (changes requested):
Few comments

pkg/nodes/list.go
elenagerman: it worked for me, all code I push was tested, at least before we start to modify it during review. But I could re-test the last code version to be sure

elenagerman: if you have any idea how we could modify it to be more stable I will be more then happy to here your suggestions.

elenagerman: @trewest please, take look on the last code version, I did some changes, looks much better and works just fine

pkg/nodes/list.go
```diff
@@ -60,3 +68,110 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+	if err != nil {
+		glog.V(100).Infof("Failed to list all nodes due to %s", err.Error())
+
+		return false, err
+	}
+
+	// Wait 5 secs in each iteration before condition function () returns true or errors or times out
+	// after availableDuration
+	err = wait.PollImmediate(backoff, timeout, func() (done bool, err error) {
+		// iterate through the ClusterOperators in the list.
```
trewest, line 89:
```suggestion
		// iterate through the Nodes in the list.
```

pkg/nodes/list.go
```diff
@@ -60,3 +68,110 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+		for _, node := range nodesList {
+			_, err = node.IsReady()
```
trewest, line 91: We need to update this to check the boolean returned from the IsReady method
```suggestion
			ready, err := node.IsReady()
			if err != nil {
				glog.V(100).Infof("Node %s has error %w", node.Object.Name, err)

				return false, err
			}
			
			if !ready {
			    glog.V(100).Infof("Node %s not Ready", node.Object.Name)
			    
		            return false, nil
		       }
```

pkg/nodes/node.go
```diff
@@ -189,6 +196,94 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+	}
+
+	err := wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		if !builder.Exists() {
+			return false, fmt.Errorf("node %s object doesn't exist", builder.Definition.Name)
+		}
+
+		for _, condition := range builder.Object.Status.Conditions {
+			if condition.Type == conditionType {
+				return condition.Status == "Unknown", nil
```
trewest, line 262: Why "Unknown" instead of "False"?

elenagerman: cause there is no False, but Unknown :)

pkg/nodes/list.go
```diff
@@ -60,3 +68,115 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+		timeout)
+
+	nodesList, err := List(apiClient, options)
+	if err != nil {
+		glog.V(100).Infof("Failed to list all nodes due to %s", err.Error())
+
+		return false, err
+	}
+
+	// Wait 5 secs in each iteration before condition function () returns true or errors or times out
```
trewest, line 86: Please update comment and remove left overs

pkg/nodes/list.go
```diff
@@ -60,3 +68,115 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+			}
+
+			if !ready {
+				glog.V(100).Infof("Node %s not Ready", node.Object.Name)
+
+				return false, nil
+			}
+		}
+
+		// Here we are always returning "false, nil" so we keep iterating throughout the stableInterval
```
trewest, line 104: Please update this comment to reflect current state

pkg/nodes/node.go
trewest: I think we should change this to WaitUntilConditionUnknown and use that for you reboot check. The False status does exist on conditions while the node is booted and reporting correctly.

pkg/nodes/list.go
```diff
@@ -60,3 +68,115 @@ func ListExternalIPv4Networks(apiClient *clients.Settings, options v1.ListOption
...
+	glog.V(100).Infof("Waiting for all nodes in the list to reboot and return to the Ready condition")
+
+	nodesList, err := List(apiClient, options)
+	if err != nil {
+		glog.V(100).Infof("Failed to list all nodes due to %s", err.Error())
+
+		return false, err
+	}
+
+	globalRebootTimeout := 2 * time.Duration(len(nodesList)) * singleNodeChangeStateTimeout
```
trewest, line 136: I think what you did here makes sense but I'm concerned that objects may change state between iterations (say a node times out waiting to be not ready and then on the next iteration it is ready. This node would be orphaned). WDYT about tracking the state of rebooted nodes similarly to the way you are tracking ready nodes? Something like this:
```
	globalStartTime := time.Now().Unix()
	readyNodes := []string{}
	rebootedNodes := []string{}
	err = wait.PollImmediate(backoff, globalRebootTimeout, func() (done bool, err error) {
		for _, node := range nodesList {
			if !slices.Contains(readyNodes, node.Object.Name) {
				ready, err := node.IsReady()
				if err != nil {
					return false, err
				}

				rebooted := slices.Contains(rebootedNodes, node.Object.Name)
				if !ready && !rebooted {
					glog.V(100).Infof("Node %s was rebooted and is starting to recover", node.Object.Name)
					
					rebootedNodes = append(rebootedNodes, node.Object.Name)
				}

				if ready && rebooted {
					glog.V(100).Infof("Node %s was successfully rebooted after: %v", time.Now().Unix() - globalStartTime)
					
					readyNodes = append(readyNodes, node.Object.Name)
				}
			}
		}

		return len(readyNodes) == len(nodesList), nil
```
I think you can also just use the timeout provided by the user for this. That way the timeout is predictable and we don't have to calculate our own timeout.

pkg/nodes/node.go
```diff
@@ -189,6 +196,94 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+	}
+
+	glog.V(100).Infof("Verify %s node availability", builder.Definition.Name)
+
+	if !builder.Exists() {
+		return false, fmt.Errorf("%s node object doesn't exist", builder.Definition.Name)
+	}
+
+	for _, condition := range builder.Object.Status.Conditions {
+		if condition.Type == "Ready" {
```
trewest, line 212: For consistency 
```suggestion
		if condition.Type == v1.NodeReady {
```

trewest (changes requested):
Just a few comment. Overall, looks good

pkg/nodes/node.go
elenagerman: I'm not so sure about it, it can confuse. Externally (oc get nodes) status called as NotReady

pkg/nodes/list.go
elenagerman: testing required, I will confirm when I'll have any result

elenagerman: tested

pkg/nodes/list.go
trewest: Please move options to the last argument in this function

pkg/nodes/node.go
```diff
@@ -189,6 +196,94 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+		}
+
+		return false, fmt.Errorf("%s node did not change it state from %s", builder.Definition.Name, conditionType)
+	})
+
+	if err == nil {
+		return nil
+	}
+
+	return fmt.Errorf("node %s did not change it state from %s until riched the timeout %s due to %w",
```
trewest, line 273:
```suggestion
	return fmt.Errorf("node %s did not change state to %s before timeout %v due to %w",
```

pkg/nodes/node.go
```diff
@@ -189,6 +196,94 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+	}
+
+	err := wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		if !builder.Exists() {
+			return false, fmt.Errorf("node %s object doesn't exist", builder.Definition.Name)
+		}
+
+		for _, condition := range builder.Object.Status.Conditions {
+			if condition.Type == conditionType {
+				return condition.Status == "True", nil
```
trewest, line 234: Please either use the isTrue const or remove it

pkg/nodes/node.go
```diff
@@ -189,6 +196,94 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+			}
+		}
+
+		return false, fmt.Errorf("%s node is not in %s state", builder.Definition.Name, conditionType)
```
trewest, line 238:
```suggestion
		return false, fmt.Errorf("the %s condition could not be found for node %s", builder.Definition.Name, conditionType)
```

pkg/nodes/node.go
```diff
@@ -189,6 +196,94 @@ func (builder *Builder) ExternalIPv4Network() (string, error) {
...
+	})
+
+	if err == nil {
+		return nil
+	}
+
+	return fmt.Errorf("%s node is not in %s state due to %w", builder.Definition.Name, conditionType, err)
```
trewest, line 245:
```suggestion
	return fmt.Errorf("%s node condition %s never became True due to %w", builder.Definition.Name, conditionType, err)
```

trewest (changes requested):
Few more minor comments. Overall looks good

trewest:
@elenagerman I think there is just one last comment - https://github.com/openshift-kni/eco-goinfra/pull/103/files#r1338621431

pkg/nodes/node.go
```diff
@@ -22,6 +29,28 @@ type Builder struct {
 	errorMsg   string
 }
 
+// NewBuilder creates new instance of node.
+func NewBuilder(apiClient *clients.Settings, name string) *Builder {
```
trewest, line 33: Do you have a use case for creating a node?

pkg/nodes/node.go
```diff
@@ -77,11 +106,79 @@ func (builder *Builder) Exists() bool {
...
 	glog.V(100).Infof("Checking if node %s exists", builder.Definition.Name)
 
+	var err error
+	_, err = builder.Get()
+
+	return err == nil || !k8serrors.IsNotFound(err)
+}
+
+// Get fetches existing node from the cluster.
+func (builder *Builder) Get() (*Builder, error) {
```
trewest, line 116: You do not need a get method for resources that provide their own get method (builder.Object, err = builder.apiClient.CoreV1Interface.Nodes().Get)

pkg/nodes/node.go
```diff
@@ -77,11 +106,79 @@ func (builder *Builder) Exists() bool {
...
+		return fmt.Errorf("can not delete node %s due to %w", builder.Definition.Name, err)
+	}
+
+	builder.Object = nil
+
+	return nil
+}
+
+// Create makes a node in the cluster and stores the created object in struct.
+func (builder *Builder) Create() (*Builder, error) {
```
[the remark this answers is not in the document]
elenagerman: possible, but not needed. Node has to be created automatically once we have bmh and mc was scaled

pkg/nodes/node.go
elenagerman: no

pkg/nodes/node.go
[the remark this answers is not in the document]
trewest: Can you remove this method? I don't think there is a need for it right now. If there is in the future we can add it along with a new builder function

kononovn (approved):
lgtm

trewest (approved):
/lgtm

PR #104, opened 2023-09-04
mco: fix mcp condition comparing definition, removing double quotes, …

elenagerman:
mco: fix mcp condition comparing definition for IsInCondition method, removing double quotes, it is const value usage


kononovn (approved):
lgtm

PR #105, opened 2023-09-05
kmm: Allowing more flexibility when building kernel-mappings.

ybettan:
In some cases, we want to build an empty kernel mapping and in some cases we want to build a kernel mapping that contains both a `regexp` as well as a `literal` in order to validate that the webhook validation is working correctly.

### DIFF
--- a/pkg/kmm/kernelmapping.go
+++ b/pkg/kmm/kernelmapping.go
@@ -21,24 +21,20 @@ type KernelMappingBuilder struct {
 // KernelMappingAdditionalOptions additional options for KernelMapping object.
 type KernelMappingAdditionalOptions func(builder *KernelMappingBuilder) (*KernelMappingBuilder, error)
 
+func NewKernelMappingBuilder() *KernelMappingBuilder {
+	glog.V(100).Infof("Initializing new empty KernelMapping parameter structure")
+
+	return &KernelMappingBuilder{
+		definition: &moduleV1Beta1.KernelMapping{},
+	}
+}
+
 // NewRegExKernelMappingBuilder creates new kernel mapping element based on regex.
 func NewRegExKernelMappingBuilder(regex string) *KernelMappingBuilder {
 	glog.V(100).Infof(
 		"Initializing new regex KernelMapping parameter structure with the following regex param: %s", regex)
 
-	builder := KernelMappingBuilder{
-		definition: &moduleV1Beta1.KernelMapping{
-			Regexp: regex,
-		},
-	}
-
-	if regex == "" {
-		glog.V(100).Infof("The regex of NewRegExKernelMappingBuilder is empty")
-
-		builder.errorMsg = "'regex' parameter can not be empty"
-	}
-
-	return &builder
+	return NewKernelMappingBuilder().WithRegexp(regex)
 }
 
 // NewLiteralKernelMappingBuilder create new kernel mapping element based on literal.
@@ -46,19 +42,7 @@ func NewLiteralKernelMappingBuilder(literal string) *KernelMappingBuilder {
 	glog.V(100).Infof(
 		"Initializing new literal KernelMapping parameter structure with following literal param: %s", literal)
 
-	builder := KernelMappingBuilder{
-		definition: &moduleV1Beta1.KernelMapping{
-			Literal: literal,
-		},
-	}
-
-	if literal == "" {
-		glog.V(100).Infof("The literal of NewLiteralKernelMappingBuilder is empty")
-
-		builder.errorMsg = "'literal' parameter can not be empty"
-	}
-
-	return &builder
+	return NewKernelMappingBuilder().WithLiteral(literal)
 }
 
 // BuildKernelMappingConfig returns kernel mapping config if error is not occur.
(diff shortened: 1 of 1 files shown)


pkg/kmm/kernelmapping.go
```diff
@@ -73,6 +57,38 @@ func (builder *KernelMappingBuilder) BuildKernelMappingConfig() (*moduleV1Beta1.
...
 }
 
+func (builder *KernelMappingBuilder) WithRegexp(regexp string) *KernelMappingBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Adding regexp %s to kernel mapping", regexp)
+
+	builder.definition.Regexp = regexp
```
kononovn, line 67: Please validate that regexp is not empty string.

pkg/kmm/kernelmapping.go
```diff
@@ -73,6 +57,38 @@ func (builder *KernelMappingBuilder) BuildKernelMappingConfig() (*moduleV1Beta1.
...
+	if regexp == "" {
+		glog.V(100).Infof("The regex of NewRegExKernelMappingBuilder is empty")
+
+		builder.errorMsg = "'regexp' parameter can not be empty"
+	}
+
+	return builder
+}
+
+func (builder *KernelMappingBuilder) WithLiteral(literal string) *KernelMappingBuilder {
```
kononovn, line 78: Please validate that literal is not empty string

pkg/kmm/kernelmapping.go
```diff
@@ -73,6 +57,38 @@ func (builder *KernelMappingBuilder) BuildKernelMappingConfig() (*moduleV1Beta1.
...
+	glog.V(100).Infof("Adding literal %s to the kernel mapping", literal)
+
+	builder.definition.Literal = literal
+
+	if literal == "" {
```
kononovn, line 83: Please move this validation before: 
builder.definition.Literal = literal
Update Literal only if builder.errorMsg == ""

pkg/kmm/kernelmapping.go
```diff
@@ -73,6 +57,38 @@ func (builder *KernelMappingBuilder) BuildKernelMappingConfig() (*moduleV1Beta1.
...
+func (builder *KernelMappingBuilder) WithRegexp(regexp string) *KernelMappingBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Adding regexp %s to kernel mapping", regexp)
+
+	builder.definition.Regexp = regexp
+
+	if regexp == "" {
```
kononovn, line 69: Please move this validation before: 
builder.definition.Regexp = regexp
Update Literal only if builder.errorMsg == ""

kononovn (changes requested):
Looks good. Please fix few comments

pkg/kmm/kernelmapping.go
ybettan: The only reason I didn't do it to begin with is because this is how the [original-function](https://github.com/openshift-kni/eco-goinfra/blob/3c31a5ef4dea4b9dd46d392f9a15dff39672e058/pkg/kmm/kernelmapping.go#L29-L39) was written and I didn't want to change the logic.

Should I still modify it? I don't see any CI in the repo and I am not sure who else uses those functions.

pkg/kmm/kernelmapping.go
ybettan: Same as https://github.com/openshift-kni/eco-goinfra/pull/105/files#r1316737580 with https://github.com/openshift-kni/eco-goinfra/blob/3c31a5ef4dea4b9dd46d392f9a15dff39672e058/pkg/kmm/kernelmapping.go#L49-L59

ybettan:
Thank you for reviewing @kononovn. I have answered the comment, please let me know if you still think the modification is required.

pkg/kmm/kernelmapping.go
kononovn: The example you showed above is for NewBuilder* functions where we store error in order to return object anyway however for mutation functions With* we don't update object if argument is invalid.

pkg/kmm/kernelmapping.go
```diff
@@ -21,44 +21,28 @@ type KernelMappingBuilder struct {
 // KernelMappingAdditionalOptions additional options for KernelMapping object.
 type KernelMappingAdditionalOptions func(builder *KernelMappingBuilder) (*KernelMappingBuilder, error)
 
+func NewKernelMappingBuilder() *KernelMappingBuilder {
```
kononovn, line 24: Can you create object with empty kernel mapping?

pkg/kmm/kernelmapping.go
```diff
@@ -21,44 +21,28 @@ type KernelMappingBuilder struct {
...
-	}
-
-	if regex == "" {
-		glog.V(100).Infof("The regex of NewRegExKernelMappingBuilder is empty")
-
-		builder.errorMsg = "'regex' parameter can not be empty"
-	}
-
-	return &builder
+	return NewKernelMappingBuilder().WithRegexp(regex)
```
kononovn, line 37: Please don't use With* function under New*Builder functions

pkg/kmm/kernelmapping.go
ybettan: Creating an empty kernelMapping means that all the values in the KernelMapping object are getting the go default value, pointers are getting `nil`, strings are getting `""` etc.
Therefore, an empty kernel mapping means that `containerImage==""`. KMM's validation webhook won't allow it.

In order to test that validation webhook, we should be able to create an empty kernelMapping in the *tests* to validate that the webhook is indeed working as expected.

pkg/kmm/kernelmapping.go
ybettan: Umm the only other options I can think of are:
1. a helper function that modifies the `kernelMapping.regexp` and then called twice: in `NewRegExKernelMappingBuilder` as well as in `WithRegexp` (to me this is the same think as the current implementation)
2. code duplication in the 2 functions

Please let me know which one you prefer, or if you have a better solution :)

kononovn:
I close this PR because we've found a better approach to make invalid Kmm object using direct access to public element of struct.

PR #106, opened 2023-09-05
bmh pkg: list handeling method was added

elenagerman:
baremetalhost.go: 
   - Pull method was replaced with Discover
   - new method added GetBmhPowerOnStatus

baremetalhosts.go:
  -  created new buider to handle bmh lists
  - WaitToBeAllInGoodOperationalStatusFor method added

### DIFF
--- a/pkg/bmh/baremetalhost.go
+++ b/pkg/bmh/baremetalhost.go
@@ -4,10 +4,10 @@ import (
 	"context"
 	"time"
 
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+
 	"github.com/golang/glog"
 	"k8s.io/apimachinery/pkg/util/wait"
-
-	goclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"fmt"
 
@@ -442,21 +442,6 @@ func (builder *BmhBuilder) Delete() (*BmhBuilder, error) {
 	return builder, nil
 }
 
-// Exists checks whether the given bmh exists.
-func (builder *BmhBuilder) Exists() bool {
-	if valid, _ := builder.validate(); !valid {
-		return false
-	}
-
-	glog.V(100).Infof("Checking if baremetalhost %s exists in namespace %s",
-		builder.Definition.Name, builder.Definition.Namespace)
-
-	var err error
-	builder.Object, err = builder.Get()
-
-	return err == nil || !k8serrors.IsNotFound(err)
-}
-
 // Get returns bmh object if found.
 func (builder *BmhBuilder) Get() (*bmhv1alpha1.BareMetalHost, error) {
 	if valid, err := builder.validate(); !valid {
@@ -477,6 +462,53 @@ func (builder *BmhBuilder) Get() (*bmhv1alpha1.BareMetalHost, error) {
 	}
 
 	return bmh, err
+}
+
+// Exists checks whether the given bmh exists.
+func (builder *BmhBuilder) Exists() bool {
+	if valid, _ := builder.validate(); !valid {
+		return false
+	}
+
+	glog.V(100).Infof("Checking if baremetalhost %s exists in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	var err error
+	builder.Object, err = builder.Get()
+
+	return err == nil || !k8serrors.IsNotFound(err)
+}
+
+// GetBmhOperationalState returns the current OperationalStatus of the bmh.
+func (builder *BmhBuilder) GetBmhOperationalState() bmhv1alpha1.OperationalStatus {
+	if valid, _ := builder.validate(); !valid {
+		return ""
+	}
+
+	glog.V(100).Infof("Pull OperationalStatus value for %s baremetalhost within %s namespace",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if !builder.Exists() {
+		return ""
+	}
+
+	return builder.Object.Status.OperationalStatus
+}
+
+// GetBmhPowerOnStatus checks BareMetalHost PowerOn status.
+func (builder *BmhBuilder) GetBmhPowerOnStatus() bool {
+	if valid, _ := builder.validate(); !valid {
+		return false
+	}
+
+	glog.V(100).Infof("Pull PoweredOn value for %s baremetalhost within %s namespace",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if !builder.Exists() {
+		return false
+	}
+
+	return builder.Object.Status.PoweredOn
 }
 
 // CreateAndWaitUntilProvisioned creates bmh object and waits until bmh is provisioned.
@@ -528,6 +560,7 @@ func (builder *BmhBuilder) WaitUntilInStatus(status bmhv1alpha1.ProvisioningStat
 	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
 		var err error
 		builder.Object, err = builder.Get()
+
 		if err != nil {
 			return false, nil
 		}
--- a/pkg/bmh/list.go
+++ b/pkg/bmh/list.go
@@ -1,0 +1,102 @@
+package bmh
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/golang/glog"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	bmhv1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+)
+
+const (
+	fiveScds time.Duration = 5 * time.Second
+)
+
+// List returns bareMetalHosts inventory in the given namespace.
+func List(apiClient *clients.Settings, nsname string) ([]*BmhBuilder, error) {
+	glog.V(100).Infof("Listing bareMetalHosts in the nsname %s", nsname)
+
+	if nsname == "" {
+		glog.V(100).Infof("bareMetalHost 'nsname' parameter can not be empty")
+
+		return nil, fmt.Errorf("failed to list bareMetalHosts, 'nsname' parameter is empty")
+	}
+
+	var bmhList bmhv1alpha1.BareMetalHostList
+	err := apiClient.List(context.Background(), &bmhList, &goclient.ListOptions{Namespace: nsname})
+
+	if err != nil {
+		glog.V(100).Infof("Failed to list bareMetalHosts in the nsname %s due to %s", nsname, err.Error())
+
+		return nil, err
+	}
+
+	var bmhObjects []*BmhBuilder
+
+	for _, baremetalhost := range bmhList.Items {
+		copiedBmh := baremetalhost
+		bmhBuilder := &BmhBuilder{
+			apiClient:  apiClient,
+			Object:     &copiedBmh,
+			Definition: &copiedBmh,
+		}
+
+		bmhObjects = append(bmhObjects, bmhBuilder)
+	}
+
+	return bmhObjects, nil
+}
+
+// WaitForAllBareMetalHostsInGoodOperationalState waits for all baremetalhosts to be in good Operational State
+// for a time duration up to the timeout.
+func WaitForAllBareMetalHostsInGoodOperationalState(apiClient *clients.Settings,
+	nsname string,
+	timeout time.Duration) (bool, error) {
+	glog.V(100).Infof("Waiting for all bareMetalHosts in %s namespace to have OK operationalStatus",
+		nsname)
+
+	bmhList, err := List(apiClient, nsname)
+	if err != nil {
+		glog.V(100).Infof("Failed to list all bareMetalHosts in the %s namespace due to %s",
+			nsname, err.Error())
+
+		return false, err
+	}
+
+	// Wait 5 secs in each iteration before condition function () returns true or errors or times out
+	// after availableDuration
+	err = wait.PollImmediate(fiveScds, timeout, func() (bool, error) {
+
+		for _, baremetalhost := range bmhList {
+			status := baremetalhost.GetBmhOperationalState()
+
+			if status != bmhv1alpha1.OperationalStatusOK {
+				glog.V(100).Infof("The %s bareMetalHost in namespace %s has an unexpected operational status: %s",
+					baremetalhost.Object.Name, baremetalhost.Object.Namespace, status)
+
+				return false, nil
+			}
+		}
+
+		return true, nil
+	})
+
+	if err == nil {
+		glog.V(100).Infof("All baremetalhosts were found in the good Operational State "+
+			"during defined timeout: %v", timeout)
+
+		return true, nil
+	}
+
+	// Here err is "timed out waiting for the condition"
+	glog.V(100).Infof("Not all baremetalhosts were found in the good Operational State "+
+		"during defined timeout: %v", timeout)
+
+	return false, err
+}


elenagerman:
@kononovn @trewest PTAL

pkg/bmh/baremetalhosts.go
```diff
@@ -0,0 +1,165 @@
...
+		glog.V(100).Infof("The namespace for baremetalhosts is empty")
+
+		builder.errorMsg = "The namespace where baremetalhosts should be found cannot be empty"
+	}
+
+	return builder
+}
+
+// DiscoverList method gets the BaremetaHosts list in cluster and stores them in the builder struct.
+func (builder *BmhListBuilder) DiscoverList() error {
```
trewest, line 52: Just call this Discover. Since it is a BmhListBuilder method, it's redundant to also use the word List in the method

pkg/bmh/baremetalhosts.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		glog.V(100).Infof("Error to list BaremetaHosts")
+
+		return err
+	}
+
+	if len(bmhList.Items) < 1 {
+		glog.V(100).Infof("Cluster doesn't have BaremetaHosts deployed, bad cluster state ...")
+
+		return fmt.Errorf("BaremetaHosts list is empty")
```
trewest, line 73: Please use lowercase characters for errors

pkg/bmh/baremetalhosts.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	glog.V(100).Infof("Wait for each BMH from the BMHList reaches "+
+		"operationalStatus: OK up to duration of %v", timeout)
+
+	isAllBmhOk := true
+
+	// Wait 5 secs in each iteration before condition function () returns true or errors or times out
+	// after availableDuration
+	err := wait.PollImmediate(fiveScds, timeout, func() (done bool, err error) {
```
trewest, line 103: Same comments as https://github.com/openshift-kni/eco-goinfra/pull/101/files#r1317940396

pkg/bmh/baremetalhost.go
```diff
@@ -363,39 +363,24 @@ func (builder *BmhBuilder) WithOptions(options ...AdditionalOptions) *BmhBuilder
 	return builder
 }
 
-// Pull pulls existing baremetalhost from cluster.
-func Pull(apiClient *clients.Settings, name, nsname string) (*BmhBuilder, error) {
```
trewest, line 367: Please restore the Pull function. This will be needed for many tests

pkg/bmh/baremetalhost.go
```diff
@@ -363,39 +363,24 @@ func (builder *BmhBuilder) WithOptions(options ...AdditionalOptions) *BmhBuilder
...
-			},
-		},
-	}
-
-	if name == "" {
-		glog.V(100).Infof("The name of the baremetalhost is empty")
-
-		builder.errorMsg = "baremetalhost 'name' cannot be empty"
+// Discover fetches existing baremetalhost from cluster.
+func (builder *BmhBuilder) Discover() error {
```
trewest, line 367: Please remove this method, we will not need it

pkg/bmh/baremetalhost.go
```diff
@@ -451,32 +436,37 @@ func (builder *BmhBuilder) Exists() bool {
...
 
-	var err error
-	builder.Object, err = builder.Get()
+	err := builder.Discover()
 
 	return err == nil || !k8serrors.IsNotFound(err)
 }
 
-// Get returns bmh object if found.
-func (builder *BmhBuilder) Get() (*bmhv1alpha1.BareMetalHost, error) {
```
trewest, line 461: Please restore the Get method. Metal3 does not provide a versioned client interface that provides a Get method

pkg/bmh/list.go
```diff
@@ -0,0 +1,107 @@
...
+	bmhv1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+)
+
+const (
+	fiveScds time.Duration = 5 * time.Second
+)
+
+// List returns bareMetalHosts inventory in the given namespace.
+func List(apiClient *clients.Settings, nsname string) ([]*BmhBuilder, error) {
```
trewest, line 22: Please add an goclient.ListOptions argument so that users can be more granular regarding the BMHs that they want to list

pkg/bmh/list.go
```diff
@@ -0,0 +1,107 @@
...
+
+		bmhObjects = append(bmhObjects, bmhBuilder)
+	}
+
+	return bmhObjects, nil
+}
+
+// WaitForAllBareMetalHostsInGoodOperationalState waits for all baremetalhosts to be in good Operational State
+// for a time duration up to the timeout.
+func WaitForAllBareMetalHostsInGoodOperationalState(apiClient *clients.Settings,
```
trewest, line 64: Please add an goclient.ListOptions argument so that users can be more granular regarding the BMHs that they want to list

pkg/bmh/list.go
```diff
@@ -0,0 +1,107 @@
...
+	nsname string,
+	timeout time.Duration) (bool, error) {
+	glog.V(100).Info("Waiting for all bareMetalHosts's operationalStatus is OK")
```
trewest, line 67: Please add namespace to glog

pkg/bmh/list.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		glog.V(100).Infof("Failed to list bareMetalHosts in the nsname %s due to %s", nsname, err.Error())
+
+		return nil, err
+	}
+
+	if len(bmhList.Items) < 1 {
+		glog.V(100).Infof("Cluster doesn't have baremetalhosts deployed, bad cluster state ...")
+
+		return nil, fmt.Errorf("BaremetaHosts list is empty")
```
trewest, line 43:
```suggestion
		return nil, fmt.Errorf("baremetaHosts list is empty")
```

trewest (changes requested):
Few comments

pkg/bmh/list.go
elenagerman: not needed, I don't see it could be useful. and, anyway, I add only methods and options I could test. If anyone will need more and will have case an environment to test it so he could add such option.

pkg/bmh/baremetalhost.go
```diff
@@ -457,26 +479,32 @@ func (builder *BmhBuilder) Exists() bool {
...
-	glog.V(100).Infof("Getting baremetalhost %s in namespace %s",
+	glog.V(100).Infof("Pull OperationalStatus value for %s baremetalhost within %s namespace",
 		builder.Definition.Name, builder.Definition.Namespace)
 
-	bmh := &bmhv1alpha1.BareMetalHost{}
-	err := builder.apiClient.Get(context.TODO(), goclient.ObjectKey{
-		Name:      builder.Definition.Name,
-		Namespace: builder.Definition.Namespace,
-	}, bmh)
+	status := builder.Object.Status.OperationalStatus
```
trewest, line 491: You just should run builder.Exists() to update builder.Object and then check the OperationalStatus

pkg/bmh/baremetalhost.go
```diff
@@ -457,26 +479,32 @@ func (builder *BmhBuilder) Exists() bool {
...
+func (builder *BmhBuilder) GetBmhPowerOnStatus() bool {
+	if valid, _ := builder.validate(); !valid {
+		return false
 	}
 
-	return bmh, err
+	glog.V(100).Infof("Pull PoweredOn value for %s baremetalhost within %s namespace",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	status := builder.Object.Status.PoweredOn
```
trewest, line 505: You just should run builder.Exists() to update builder.Object and then check PoweredOn

pkg/bmh/baremetalhost.go
```diff
@@ -526,8 +554,7 @@ func (builder *BmhBuilder) WaitUntilInStatus(status bmhv1alpha1.ProvisioningStat
 	}
 
 	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
-		var err error
-		builder.Object, err = builder.Get()
```
trewest, line 530: I think this line is more accurate. You need to update builder.Object before checking .Status.Provisioning.State, otherwise you will just be reading the same value each time

trewest: Or you can just use builder.Exists() and also return an error if it doesn't exist

pkg/bmh/list.go
trewest: It's useful because some folks may need to only get a subset of BMHs based on things such as LabelSelectors or FieldSelectors. Typically I agree that you should only add what you need but in this case changing it later what cause you to have to refactor any tests that use it. Please reference this [function](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/pod/list.go#L76) as an example.

elenagerman: Maybe I was not clear; let me explain it again.
There is no use case for the bmh with options. After 4 years with ocp-edge-auto, I can tell you we don't need it.
In addition, it will be a bit complicated to implement; on the primary List method, we are already using a goclient.ListOptions to filter bmh list by namespace (at least for me on golang, I don't understand how to do so)
The primary usage for a bmh object is to work directly with a specific bmh object, a single one. The list is mainly used to overview all existing bmh, useful in case of scale up/down, etcd, replacement to choose random bmh. You could look at the bmh object to understand that there is nothing to filter out except the namespace.

pkg/bmh/list.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return bmhObjects, nil
+}
+
+// WaitForAllBareMetalHostsInGoodOperationalState waits for all baremetalhosts to be in good Operational State
+// for a time duration up to the timeout.
+func WaitForAllBareMetalHostsInGoodOperationalState(apiClient *clients.Settings,
+	nsname string,
+	listOptions metav1.ListOptions,
```
trewest, line 67: This argument is unused

pkg/bmh/baremetalhost.go
```diff
@@ -457,26 +444,40 @@ func (builder *BmhBuilder) Exists() bool {
...
 
-	return bmh, err
+	glog.V(100).Infof("Pull PoweredOn value for %s baremetalhost within %s namespace",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if !builder.Exists() {
+		return false
+	}
+
+	status := builder.Object.Status.PoweredOn
```
trewest, line 478: This can be simplified
```suggestion
	return builder.Object.Status.PoweredOn
```

pkg/bmh/baremetalhost.go
```diff
@@ -457,26 +444,40 @@ func (builder *BmhBuilder) Exists() bool {
...
-		Name:      builder.Definition.Name,
-		Namespace: builder.Definition.Namespace,
-	}, bmh)
+	if !builder.Exists() {
+		return ""
+	}
 
-	if err != nil {
-		return nil, err
+	status := builder.Object.Status.OperationalStatus
```
trewest, line 460: This can be simplified
```suggestion
	return builder.Object.Status.OperationalStatus
```

pkg/bmh/baremetalhost.go
trewest: Hey I still see that pull is removed

kubeconfig
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
```
trewest, line 1: Seems to be added by accident

pkg/bmh/baremetalhost.go
elenagerman: I restored it now, but pay attention that I have Get method

pkg/bmh/list.go
elenagerman: yep, thanks

pkg/bmh/list.go
```diff
@@ -0,0 +1,108 @@
...
+	var bmhList bmhv1alpha1.BareMetalHostList
+	err := apiClient.List(context.Background(), &bmhList, &goclient.ListOptions{Namespace: nsname})
+
+	if err != nil {
+		glog.V(100).Infof("Failed to list bareMetalHosts in the nsname %s due to %s", nsname, err.Error())
+
+		return nil, err
+	}
+
+	if len(bmhList.Items) < 1 {
```
trewest, line 40: This isn't required. Just let the function return an empty list and let the user determine if the cluster is in a bad state or not.

pkg/bmh/list.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// Wait 5 secs in each iteration before condition function () returns true or errors or times out
+	// after availableDuration
+	err = wait.PollImmediate(fiveScds, timeout, func() (bool, error) {
+
+		for _, baremetalhost := range bmhList {
+			status := baremetalhost.GetBmhOperationalState()
+
+			if status != bmhv1alpha1.OperationalStatusOK {
+				glog.V(100).Infof("The %s bareMetalHost not available; status: %s",
```
trewest, line 86:
```suggestion
				glog.V(100).Infof("The %s bareMetalHost in namespace %s has an unexpected operational status: %s",
```

pkg/bmh/baremetalhost.go
```diff
@@ -457,26 +479,36 @@ func (builder *BmhBuilder) Exists() bool {
 	return err == nil || !k8serrors.IsNotFound(err)
 }
 
-// Get returns bmh object if found.
-func (builder *BmhBuilder) Get() (*bmhv1alpha1.BareMetalHost, error) {
-	if valid, err := builder.validate(); !valid {
-		return nil, err
+// GetBmhOperationalState checks whether the given bmh exists.
```
trewest, line 482:
```suggestion
// GetBmhOperationalState returns the current OperationalStatus of the bmh.
```

trewest (changes requested):
Overall looks good. Just a few minor comments

pkg/bmh/list.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return bmhObjects, nil
+}
+
+// WaitForAllBareMetalHostsInGoodOperationalState waits for all baremetalhosts to be in good Operational State
+// for a time duration up to the timeout.
+func WaitForAllBareMetalHostsInGoodOperationalState(apiClient *clients.Settings,
+	nsname string,
+	timeout time.Duration) (bool, error) {
+	glog.V(100).Infof("Waiting for all bareMetalHosts's in %s namespace operationalStatus is OK",
```
trewest, line 61:
```suggestion
	glog.V(100).Infof("Waiting for all bareMetalHosts in %s namespace to have OK operationalStatus",
```

pkg/bmh/list.go
```diff
@@ -0,0 +1,102 @@
...
+	// Wait 5 secs in each iteration before condition function () returns true or errors or times out
+	// after availableDuration
+	err = wait.PollImmediate(fiveScds, timeout, func() (bool, error) {
+
+		for _, baremetalhost := range bmhList {
+			status := baremetalhost.GetBmhOperationalState()
+
+			if status != bmhv1alpha1.OperationalStatusOK {
+				glog.V(100).Infof("The %s bareMetalHost in namespace %s has an unexpected operational status: %s",
+					baremetalhost.Object.Name, status)
```
trewest, line 81:
```suggestion
					baremetalhost.Object.Name, baremetalhost.Object.Namespace, status)
```

trewest (changes requested):
Just a few minor comments, otherwise looks good

kononovn (approved):
lgtm

trewest (approved):
/lgtm

PR #111, opened 2023-09-07
kmm: adding an `Update` function for modules.

ybettan:
This is needed in order to test a module edit in KMM and make sure the old kernel-module is unloaded in favor of a new kernel-module which is being loaded.


cdvultur (approved):
/lgtm

Seems that we missed this one..

kononovn (approved):
lgtm

PR #112, opened 2023-09-08
Added networkpolicy pkg

elenagerman:


elenagerman:
@kononovn @trewest PTAL

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,292 @@
...
+		glog.V(100).Infof("The namespace of the networkPolicy is empty")
+
+		builder.errorMsg = "The networkPolicy 'namespace' cannot be empty"
+	}
+
+	return builder
+}
+
+// WithNamespaceIngressRule ingress rule for the networkPolicy.
+func (builder *NetworkPolicyBuilder) WithNamespaceIngressRule(
```
kononovn, line 61: Please re-factor this function according to:

```
// WithNamespaceIngressRule ingress rule for the networkPolicy.
func (builder *NetworkPolicyBuilder) WithNamespaceIngressRule(namespaceIngressMatchLabels map[string]string,
	podIngressMatchLabels map[string]string) *NetworkPolicyBuilder {
		builder.errorMsg = err.Error()

		return builder
	}

	if builder.errorMsg != "" {
		return builder
	}

	glog.V(100).Infof(
		"Applying Ingress rule to networkPolicy %s in namespace %s",

	if namespaceIngressMatchLabels == nil && podIngressMatchLabels == nil {
		glog.V(100).Infof("At least one type of the selector for NetworkPolicy ingress rule should be defined")

		builder.errorMsg = "Both namespaceIngressMatchLabels and podIngressMatchLabels parameters are empty maps"

		return builder
	}

	var IngressRule netv1.NetworkPolicyPeer
	if namespaceIngressMatchLabels != nil {
		glog.V(100).Infof(
			"Applying Ingress rule with namespaceIngressMatchLabels %v parameter to networkPolicy %s in namespace %s",
			namespaceIngressMatchLabels, builder.Definition.Name, builder.Definition.Namespace)

		IngressRule.NamespaceSelector = &metav1.LabelSelector{
			MatchLabels: namespaceIngressMatchLabels,
		}
	}

	if podIngressMatchLabels != nil {
		glog.V(100).Infof(
			"Applying Ingress rule with podIngressMatchLabels %v parameter to networkPolicy %s in namespace %s",
			podIngressMatchLabels, builder.Definition.Name, builder.Definition.Namespace)

		IngressRule.PodSelector = &metav1.LabelSelector{
			MatchLabels: podIngressMatchLabels,
		}
	}

	if builder.Definition.Spec.Ingress == nil {
		builder.Definition.Spec.Ingress = make([]netv1.NetworkPolicyIngressRule, 1)
	}

	ingressRule := netv1.NetworkPolicyIngressRule{}
	ingressRule.From = append(ingressRule.From, IngressRule)

	builder.Definition.Spec.Ingress = append(builder.Definition.Spec.Ingress, ingressRule)
	return builder
}
```

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,300 @@
...
+		return builder
+	}
+
+	builder.Definition.Spec.PolicyTypes = append(builder.Definition.Spec.PolicyTypes, policyType)
+
+	return builder
+}
+
+// WithPodSelector add podSelector to the networkPolicy.
+func (builder *NetworkPolicyBuilder) WithPodSelector(podSelectorMatchLabels map[string]string) *NetworkPolicyBuilder {
```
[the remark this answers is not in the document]
elenagerman: I will, but I don't think you a right, few things:

- I don't understand why to remove 'return builder' from the "if podSelectorMatchLabels == nil" block and than build another if block for return, if this is only the case where return with error could be executed???
- Why put getting and assigning a value in one line? in case of a fall, this makes it difficult to debug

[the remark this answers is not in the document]
kononovn: Regarding the first point:
Anyway we have this block if builder.errorMsg != ""  before we actually update struct so having additional return is simply duplication

Regarding the second point. I don't think we need have standalone variable that is used only once.

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,181 @@
...
+	ingressRule.From = append(ingressRule.From, peerRule)
+
+	builder.Definition.Spec.Ingress = append(builder.Definition.Spec.Ingress, ingressRule)
+
+	return builder
+}
+
+// validate will check that the builder and builder definition are properly initialized before
+// accessing any member fields.
+func (builder *NetworkPolicyBuilder) validate() (bool, error) {
```
kononovn, line 119: Please move private function to the end of the file

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,181 @@
...
+		glog.V(100).Infof("The %s builder has error message: %s", resourceCRD, builder.errorMsg)
+
+		return false, fmt.Errorf(builder.errorMsg)
+	}
+
+	return true, nil
+}
+
+// Create makes a networkPolicy in cluster and stores the created object in struct.
+func (builder *NetworkPolicyBuilder) Create() (*NetworkPolicyBuilder, error) {
```
kononovn, line 150: Can you please add Delete and Update function to this package?

kononovn (changes requested):
Please fix 2 more comments.

pkg/networkpolicy/networkpolicy.go
elenagerman: it is

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,311 @@
...
+	Object *netv1.NetworkPolicy
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
+	// errorMsg is processed before NetworkPolicy object is created.
+	errorMsg string
+}
+
+// NewNetworkPolicyBuilder method creates new instance of builder.
+func NewNetworkPolicyBuilder(
+	apiClient *clients.Settings, name, nsname string) *NetworkPolicyBuilder {
```
kononovn, line 29: Parameters can be in the same line with function.

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,311 @@
...
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the networkPolicy is empty")
+
+		builder.errorMsg = "The networkPolicy 'namespace' cannot be empty"
+	}
+
+	return builder
+}
+
+// WithNamespaceIngressRule ingress rule for the networkPolicy.
```
kononovn, line 58: Please change comment to>
// WithNamespaceIngressRule applies ingress rule for the networkPolicy.

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,311 @@
...
+func (builder *NetworkPolicyBuilder) WithNamespaceIngressRule(namespaceIngressMatchLabels map[string]string,
+	podIngressMatchLabels map[string]string) *NetworkPolicyBuilder {
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
+
+		return builder
```
kononovn, line 64: You can remove this return because you check it on line 67

kononovn, line 64: You can remove this return because you check the same on line 67

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,311 @@
...
+// WithPolicyType add policyType to the networkPolicy.
+func (builder *NetworkPolicyBuilder) WithPolicyType(policyType netv1.PolicyType) *NetworkPolicyBuilder {
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
+
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating networkPolicy %s with the policyType defined: %d", builder.Definition.Name)
```
kononovn, line 126: Please fix log message. One parameter is missing >
glog.V(100).Infof(
		"Creating networkPolicy %s with the policyType: %v", builder.Definition.Name, policyType)

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,311 @@
...
+
+// WithPodSelector add podSelector to the networkPolicy.
+func (builder *NetworkPolicyBuilder) WithPodSelector(podSelectorMatchLabels map[string]string) *NetworkPolicyBuilder {
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
+
+		return builder
+	}
+
+	glog.V(100).Infof(
```
kononovn, line 149: Please add podSelectorMatchLabels parameter to the log message using %v formatter

kononovn (changes requested):
Few cosmetic changes are required

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,314 @@
...
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating networkPolicy %s with the policyType defined: %d", builder.Definition.Name, policyType)
```
kononovn, line 127: Please change formatter  to %v instead of %d

kononovn (approved):
/lgtm

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+
+	return builder
+}
+
+// WithNamespaceIngressRule applies ingress rule for the networkPolicy.
+func (builder *NetworkPolicyBuilder) WithNamespaceIngressRule(
+	namespaceIngressMatchLabels map[string]string,
+	podIngressMatchLabels map[string]string) *NetworkPolicyBuilder {
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
```
trewest, line 62: We usually just return err here. builder.errorMsg will already be set in the validate method if something goes wrong

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+		builder.errorMsg = "Both namespaceIngressMatchLabels and podIngressMatchLabels parameters are empty maps"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	var peerRule netv1.NetworkPolicyPeer
+
+	if namespaceIngressMatchLabels != nil {
```
trewest, line 85: Use len(namespaceIngressMatchLabels) != 0

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+		glog.V(100).Infof(
+			"Applying Ingress rule with podIngressMatchLabels %v parameter to networkPolicy %s in namespace %s",
+			podIngressMatchLabels, builder.Definition.Name, builder.Definition.Namespace)
+
+		peerRule.PodSelector = &metav1.LabelSelector{
+			MatchLabels: podIngressMatchLabels,
+		}
+	}
+
+	if builder.Definition.Spec.Ingress == nil {
```
trewest, line 105: I think this will already be initialized. Can you check if you need it?

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating networkPolicy %s with the policyType defined: %v", builder.Definition.Name, policyType)
+
+	if policyType == "" {
+		glog.V(100).Infof("The policyType value have to be provided")
+
+		builder.errorMsg = "The policyType is empty sting"
```
trewest, line 133:
```suggestion
		builder.errorMsg = "The policyType is an empty string"
```

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating networkPolicy %s with the policyType defined: %v", builder.Definition.Name, policyType)
+
+	if policyType == "" {
+		glog.V(100).Infof("The policyType value have to be provided")
```
trewest, line 131:
```suggestion
		glog.V(100).Infof("The policyType value has to be provided")
```

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating networkPolicy %s with the policyType defined: %v", builder.Definition.Name, policyType)
```
trewest, line 128: Please include the namespace in the glog

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating networkPolicy %s with podSelector defined: %v",
```
trewest, line 156: Please include namespace in glog

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the networkPolicy is empty")
+
+		builder.errorMsg = "networkPolicy 'namespace' cannot be empty"
+	}
+
+	if builder.errorMsg != "" {
+		return nil, fmt.Errorf("faield to pull networkPolicy object due to the following error: %s",
```
trewest, line 203:
```suggestion
		return nil, fmt.Errorf("failed to pull networkPolicy object due to the following error: %s",
```

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+	return err == nil || !k8serrors.IsNotFound(err)
+}
+
+// Delete removes a networkPolicy object from a cluster.
+func (builder *NetworkPolicyBuilder) Delete() error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof("Deleting the networkPolicy object %s",
```
trewest, line 259: Please include namespace in glog

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+		return fmt.Errorf("NetworkPolicy cannot be deleted because it does not exist")
+	}
+
+	err := builder.apiClient.NetworkPolicies(builder.Definition.Namespace).Delete(
+		context.TODO(), builder.Definition.Name, metav1.DeleteOptions{})
+
+	if err != nil {
+		return fmt.Errorf("cannot delete NetworkPolicy: %w", err)
+	}
+
```
trewest, line 272: set builder.Object = nil

trewest (changes requested):
A few comments

pkg/networkpolicy/networkpolicy.go
kononovn: It's duplication since we have exactly the same logic on line 65

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+	builder.Definition.Spec.Ingress = append(builder.Definition.Spec.Ingress, ingressRule)
+
+	return builder
+}
+
+// WithPolicyType add policyType to the networkPolicy.
+func (builder *NetworkPolicyBuilder) WithPolicyType(policyType netv1.PolicyType) *NetworkPolicyBuilder {
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
```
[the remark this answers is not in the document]
kononovn: We have the same logic on line 123. If we add additional return here it's simply duplication

pkg/networkpolicy/networkpolicy.go
trewest: Right, line 65 should be removed and the return should just be done here. That's how validate is used in the other methods

pkg/networkpolicy/networkpolicy.go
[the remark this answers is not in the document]
trewest: Right, line 123 should be removed and the return should just be done here. That's how validate is used in the other methods

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,321 @@
...
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Applying Ingress rule to networkPolicy %s in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if namespaceIngressMatchLabels == nil && podIngressMatchLabels == nil {
```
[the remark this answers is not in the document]
kononovn: Ok I didn't know that empty map is **!= nil**. The issue is that function argument could be either empty map or nil. In this case we need to use more complex check:

```
func doesMapEmpty(mapper map[string]string) bool {
	return isNil(mapper) || isEmpty(mapper)
}

```
@trewest WDYT ?

pkg/networkpolicy/networkpolicy.go
kononovn: Ok I got it.
```
	builder.errorMsg = err.Error()

        return bulder
}

```
And get rid of next condition on line 65

pkg/networkpolicy/networkpolicy.go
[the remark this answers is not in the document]
kononovn: Ok I got it.
```
	builder.errorMsg = err.Error()

        return bulder
}

```
And get rid of next condition on line 123

pkg/networkpolicy/networkpolicy.go
elenagerman: I could not return err here since only the object I return from this method is NetworkPolicyBuilder builder

elenagerman: ok, now I see what you mean

pkg/networkpolicy/networkpolicy.go
elenagerman: as far as I remember I had problems with this

trewest: #L133 should give you the same issue if that is the case.  Please verify if you need it. If so, align WithPolicyType with this functionality

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	glog.V(100).Infof(
+		"Creating networkPolicy %s in %s namespace with podSelector defined: %v",
+		builder.Definition.Name, builder.Definition.Namespace, podSelectorMatchLabels)
+
+	if len(podSelectorMatchLabels) == 0 {
+		glog.V(100).Infof("The podSelector could not be empty")
+
+		builder.errorMsg = "The podSelector is empty sting"
```
trewest, line 151:
```suggestion
		builder.errorMsg = "The podSelector is an empty string"
```

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,312 @@
...
+func (builder *NetworkPolicyBuilder) Delete() error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof("Deleting the networkPolicy object %s from %s namespace",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if !builder.Exists() {
+		return fmt.Errorf("NetworkPolicy cannot be deleted because it does not exist")
```
trewest, line 251:
```suggestion
		return fmt.Errorf("networkPolicy cannot be deleted because it does not exist")
```

trewest (changes requested):
Three small comments

pkg/networkpolicy/networkpolicy.go
elenagerman: yea, wit was there, seems to be missed on some stage, too much changes were done since it was originally opened

pkg/networkpolicy/networkpolicy.go
```diff
@@ -0,0 +1,316 @@
...
+
+	if !builder.Exists() {
+		return fmt.Errorf("networkPolicy cannot be deleted because it does not exist")
+	}
+
+	err := builder.apiClient.NetworkPolicies(builder.Definition.Namespace).Delete(
+		context.TODO(), builder.Definition.Name, metav1.DeleteOptions{})
+
+	if err != nil {
+		builder.Object = nil
```
trewest, line 262: builder.Object should be set to nil under this block.

trewest (approved):
/lgtm

kononovn (approved):
/lgtm

Issue #118, opened 2023-09-08
Refactor list functions of nodes pkg.


kononovn:
Refactor list functions. Use https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/pod/list.go as example

PR #124, opened 2023-09-12
Refactor nodes pkg list function

trewest:
Updates nodes pkg to comply with List standard.

Removes nodes.go

Adds:
- list.go 
  - func List 
  - func ListExternalIPv4Networks

Closes https://github.com/openshift-kni/eco-goinfra/issues/118


kononovn (approved):
/lgtm

PR #128, opened 2023-09-15
kubeletconfig handler added to the mco package

elenagerman:


trewest:
Duplicates https://github.com/openshift-kni/eco-goinfra/pull/138

pkg/mco/kubeletconfig.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	mcv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+)
+
+// Builder provides struct for KubeletConfig Object which contains connection to cluster
+// and KubeletConfig definitions.
+type Builder struct {
```
trewest, line 20: Since this is in the mco package with other structs. we should call it something like KubeletConfigBuilder

pkg/mco/kubeletconfig.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	if memory == "" {
+		glog.V(100).Infof("The memory can't be empty")
+
+		builder.errorMsg = "'memory' cannot be empty"
+
+		return builder
+	}
+
```
trewest, line 194: Instead of returning builder twice above, just check if errorMsg != "" here and return if so

pkg/mco/kubeletconfig.go
```diff
@@ -0,0 +1,268 @@
...
+	systemReservedKubeletConfiguration := &kubeletconfigv1beta1.KubeletConfiguration{
+		SystemReserved: map[string]string{
+			cpu:    cpu,
+			memory: memory,
+		},
+	}
+
+	builder.Definition.Spec.KubeletConfig.Object = systemReservedKubeletConfiguration
```
trewest, line 210: I see that you are initializing `builder.Definition.Spec.KubeletConfig.Raw` above but then don't use it. Is that intentional?

trewest (changes requested):
Just a few comments

pkg/mco/kubeletconfig.go
elenagerman: it was, you did not react and I changed it half an our ago

pkg/mco/kubeletconfig.go
elenagerman: I don't remember already if I had an issu with no this initialization.
it was 3 weeks ago and since then a lot of code was written, I remove it now, it can be just unnecessary leftover.
maximum I'll fix it after

pkg/mco/kubeletconfig.go
```diff
@@ -0,0 +1,264 @@
...
+	errorMsg string
+}
+
+// AdditionalOptions for kubeletconfig object.
+type AdditionalOptions func(builder *KubeletConfigBuilder) (*KubeletConfigBuilder, error)
+
+// NewBuilder provides struct for KubeletConfig object which contains connection to cluster
+// and KubeletConfig definition.
+func NewBuilder(apiClient *clients.Settings, name string) *KubeletConfigBuilder {
+	glog.V(100).Infof("Initializing new KubeletConfigBuilder structure with following params: %s", name)
```
achuzhoy, line 37: replace "with following params" with " with the name"

pkg/mco/kubeletconfig.go
```diff
@@ -0,0 +1,264 @@
...
+	return builder
+}
+
+// WithSystemReserved redefines kubeletconfig definition with the given systemreserved fields.
+func (builder *KubeletConfigBuilder) WithSystemReserved(cpu, memory string) *KubeletConfigBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Setting up %s cpu and %s memory to the %s kubeletconfig",
```
achuzhoy, line 176: nit: Based on the expected entries (i.e. cpu=200m,memory=150G) the log should be better rewritten as:
"Setting cpu=%s and memory=%s in the kubeletconfig definition"

achuzhoy (approved):

pkg/mco/kubeletconfig.go
elenagerman: sure

trewest (dismissed):
/lgtm

elenagerman:
@achuzhoy do you have any comments regarding this one?

PR #131, opened 2023-09-19
Add WithHostPid func to pod.go

shaior:


pkg/pod/pod.go
```diff
@@ -724,6 +724,25 @@ func (builder *Builder) WithHostNetwork() *Builder {
 	return builder
 }
 
+// WithHostPid allows a pod to have access to the host process ID namespace.
+func (builder *Builder) WithHostPid() *Builder {
```
kononovn, line 728: Add boolean flag as a parameter. May be in some cases you'll need explicitly false flag.

pkg/pod/pod.go
```diff
@@ -724,6 +724,25 @@ func (builder *Builder) WithHostNetwork() *Builder {
...
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Applying HostPID flag to pod's %s configuration", builder.Definition.Name)
```
achuzhoy, line 733: nit: if the pod name is "sun" - then this reads:
"Applying HostPID flag to pod's sun configuration"  which can be confusing.

Suggest to change:
"Applying HostPID flag to the configuration of pod %s"

trewest: Please also include the namespace that the pod is running in

pkg/pod/pod.go
```diff
@@ -724,6 +724,26 @@ func (builder *Builder) WithHostNetwork() *Builder {
 	return builder
 }
 
+// WithHostPid configures a pod's access to the host process ID namespace based on a boolean parameter.
+func (builder *Builder) WithHostPid(flag bool) *Builder {
```
kononovn, line 728: Please re-name argument from "flag" to simple "hostPid"

shaior: done.

kononovn (approved):
/lgtm

pkg/pod/pod.go
```diff
@@ -724,6 +724,26 @@ func (builder *Builder) WithHostNetwork() *Builder {
 	return builder
 }
 
+// WithHostPid configures a pod's access to the host process ID namespace based on a boolean parameter.
+func (builder *Builder) WithHostPid(hostPid bool) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Applying HostPID flag to the configuration of pod: %s in namespace: %s:",
```
trewest, line 733: nit
```suggestion
	glog.V(100).Infof("Applying HostPID flag to the configuration of pod: %s in namespace: %s",
```

trewest (changes requested):
Small nit, otherwise lgtm

trewest (approved):
/lgtm

PR #134, opened 2023-09-20
Refactor mcp pkg list functions

kononovn:


pkg/mco/mcpList.go
```diff
@@ -8,83 +8,73 @@ import (
...
-
-// NewMCPListBuilder method creates new instance of MCPListBuilder.
-func NewMCPListBuilder(apiClient *clients.Settings, mcpLabelSelector ...map[string]string) *MCPListBuilder {
-	glog.V(100).Infof(
-		"Initializing new MCPListBuilder structure with the following params: %v", mcpLabelSelector)
-
-	builder := &MCPListBuilder{
-		apiClient: apiClient,
-	}
+func ListMCP(apiClient *clients.Settings, mcpLabelSelector ...map[string]string) ([]*MCPBuilder, error) {
```
trewest, line 16: I think you should just have options v1.ListOptions as an argument rather than mcpLabelSelector. This is what we currently do in other pkgs

pkg/mco/mcpList.go
```diff
@@ -8,83 +8,73 @@ import (
...
-		return err
+		mcpObjects = append(mcpObjects, mcpBuilder)
 	}
 
-	if len(mcpList.Items) < 1 {
-		glog.V(100).Infof("Cluster doesn't have MachineConfigPools installed ")
+	return mcpObjects, nil
+}
+
+func GetByLabelFromMCPList(apiClient *clients.Settings, mcpLabel string) (*MCPBuilder, error) {
```
trewest, line 49: This function seems to not only check labels but also expressions. Maybe it would be better to pass in a label selector and evaluate the machineconfigpool's Spec.MachineConfigSelector.MatchExpressions and Spec.MachineConfigSelector.MatchLabels values against it?

trewest (changes requested):
Couple of initial comments

pkg/mco/mcplist.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func ListMCP(apiClient *clients.Settings, listOptions metav1.ListOptions) ([]*MCPBuilder, error) {
+	glog.V(100).Infof("Listing all MCP resources with the options %v\"", listOptions)
```
trewest, line 16: Did you mean to leave this quote in?
```suggestion
	glog.V(100).Infof("Listing all MCP resources with the options %v", listOptions)
```

pkg/mco/mcplist.go
```diff
@@ -0,0 +1,138 @@
...
+			if label == mcpLabel {
+				return mcp, nil
+			}
+		}
+	}
+
+	return nil, fmt.Errorf("cannot find MachineConfigPool that targets machineConfig with label: %s", mcpLabel)
+}
+
+func ListMCPWaitToBeStableFor(apiClient *clients.Settings, stableDuration, timeout time.Duration) error {
```
trewest, line 70: I think nesting two wait.PollImmediate functions and using two timeout arguments is confusing and can lead to unexpected behavior. I would prefer to just have one timeout and one wait.PollImmediate function @kononovn  WDYT?

kononovn: It's impossible for list. Please check function logic carefully. We do need nested timeout because we iterate over list and if even one MCP is not stable we reset second timer.

trewest (approved):
/lgtm

PR #142, opened 2023-09-25
console pkg added

elenagerman:


pkg/console/console.go
```diff
@@ -0,0 +1,109 @@
...
+				Name: "cluster",
+			},
+			Spec: v1.ConsoleSpec{
+				Authentication: v1.ConsoleAuthentication{},
+			},
+		},
+	}
+
+	if valid, err := builder.validate(); !valid {
+		glog.V(100).Info("Bad cluster state, failed to pull cluster console. Object doesn't exist")
```
achuzhoy, line 60: Let's remove the assumption about the cluster state.

pkg/console/console.go
```diff
@@ -0,0 +1,109 @@
...
+	// console definition, used to create the pod object.
+	Definition *v1.Console
+	// Created console object.
+	Object *v1.Console
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
+}
+
+// AdditionalOptions additional options for console object.
+type AdditionalOptions func(builder *Builder) (*Builder, error)
```
achuzhoy, line 25: Don't see where it's used here.

pkg/console/console.go
elenagerman: sure

pkg/console/console.go
elenagerman: finally it does not, I'll remove it on this stage and if it will be required we'll add it back later

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	v1 "github.com/openshift/api/config/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Builder provides a struct for console object from the cluster and a console definition.
+type Builder struct {
+	// console definition, used to create the pod object.
```
kononovn, line 16: Please use capital letter. Console

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+	Definition *v1.Console
+	// Created console object.
+	Object *v1.Console
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
+}
+
+// NewBuilder creates a new instance of Builder.
+func NewBuilder(apiClient *clients.Settings) *Builder {
```
kononovn, line 25: Please add argument: "name string". If later they'll change default name from cluster to something else we don't need to refactor our code.

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+	glog.V(100).Info("Initializing new console structure")
+
+	builder := Builder{
+		apiClient: apiClient,
+		Definition: &v1.Console{
+			ObjectMeta: metaV1.ObjectMeta{
+				Name: "cluster",
```
kononovn, line 32: Hardcoded **cluster** should be an argument.

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+			},
+		},
+	}
+
+	return &builder
+}
+
+// Pull loads an existing console into the Builder struct.
+func Pull(apiClient *clients.Settings) (*Builder, error) {
```
kononovn, line 41: Same here. Please add argument name string to function's argument.

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+	Definition *v1.Console
+	// Created console object.
+	Object *v1.Console
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
+}
+
+// NewBuilder creates a new instance of Builder.
+func NewBuilder(apiClient *clients.Settings) *Builder {
+	glog.V(100).Info("Initializing new console structure")
```
kononovn, line 26: Update log to use new argument name.

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+			},
+		},
+	}
+
+	return &builder
+}
+
+// Pull loads an existing console into the Builder struct.
+func Pull(apiClient *clients.Settings) (*Builder, error) {
+	glog.V(100).Infof("Pulling cluster console")
```
kononovn, line 42: Update log to use new argument name.

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+			Spec: v1.ConsoleSpec{
+				Authentication: v1.ConsoleAuthentication{},
+			},
+		},
+	}
+
+	if valid, err := builder.validate(); !valid {
+		glog.V(100).Info("failed to pull cluster console. Object doesn't exist")
+
+		return nil, fmt.Errorf("cluster console object doesn't exist; %w", err)
```
kononovn, line 59: Update log to use new argument name.

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+	return &builder, nil
+}
+
+// Update renovates the existing cluster console object with cluster console definition in builder.
+func (builder *Builder) Update() (*Builder, error) {
+	if valid, err := builder.validate(); !valid {
+		return builder, err
+	}
+
+	glog.V(100).Info("Updating cluster console")
```
kononovn, line 73: Please use name of the object in the log.

pkg/console/console.go
```diff
@@ -0,0 +1,106 @@
...
+
+	var err error
+	builder.Object, err = builder.apiClient.Consoles().Update(context.Background(), builder.Definition,
+		metaV1.UpdateOptions{})
+
+	return builder, err
+}
+
```
kononovn, line 81: Please add common functions. Create/Delete/Exist to this package.

kononovn (changes requested):
Looks good. Just few comments to fix

pkg/console/console.go
elenagerman: there is no usage for Create/Delete in case of console. Maybe only Exists

pkg/console/console.go
```diff
@@ -0,0 +1,171 @@
...
+	return &builder
+}
+
+// Pull loads an existing console into the Builder struct.
+func Pull(apiClient *clients.Settings, name string) (*Builder, error) {
+	if name == "" {
+		name = "cluster"
+	}
+
+	glog.V(100).Infof("Pulling cluster console %s", name)
```
kononovn, line 52: Please add **structure** same as you have in NewBuilder

pkg/console/console.go
kononovn: It is needed for future interface.

pkg/console/console.go
```diff
@@ -0,0 +1,171 @@
...
+	Definition *v1.Console
+	// Created console object.
+	Object *v1.Console
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
+}
+
+// NewBuilder creates a new instance of Builder.
+func NewBuilder(apiClient *clients.Settings, name string) *Builder {
+	if name == "" {
```
kononovn, line 28: Code inconsistency comparing to other packages. 
Can you move this check after builder init. Put it on line 42.
Please use our regular handler:

```
	if name == "" {
		glog.V(100).Infof("The name of the Console is empty")

		builder.errorMsg = "MachineConfig 'name' cannot be empty"
	}
```

pkg/console/console.go
```diff
@@ -0,0 +1,171 @@
...
+			},
+		},
+	}
+
+	return &builder
+}
+
+// Pull loads an existing console into the Builder struct.
+func Pull(apiClient *clients.Settings, name string) (*Builder, error) {
+	if name == "" {
```
kononovn, line 48: Code inconsistency comparing to other packages. 
Can you move this check after builder init. Put it on line 65.
Please use our regular handler:

```
	if name == "" {
		glog.V(100).Infof("The name of the Console is empty")

		builder.errorMsg = "Console 'name' cannot be empty"
	}
```

pkg/console/console.go
```diff
@@ -0,0 +1,171 @@
...
+			},
+		},
+	}
+
+	if valid, err := builder.validate(); !valid {
+		glog.V(100).Info("failed to pull cluster console %s. Object doesn't exist", name)
+
+		return nil, fmt.Errorf("cluster console %s object doesn't exist; %w", name, err)
+	}
+
```
kononovn, line 71: Please add Exists function call here:

```
 	if !builder.Exists() {
		return nil, fmt.Errorf("Console object %s doesn't exist", name)
	}

	builder.Definition = builder.Objec
```t

**Review by kononovn — CHANGES_REQUESTED** [review]
Please fix few comments

File: pkg/console/console.go (hunk 13 above)
Comment (elenagerman) [reply pkg/console/console.go:28]: I don't understand why you do not accept different from the error reaction on the empty field when only one single and same for all deployment option is available. I'm still insist we do not need name parameter at all in this case.

File: pkg/console/console.go [hunk 16]
```diff
@@ -0,0 +1,176 @@
...
+			ObjectMeta: metaV1.ObjectMeta{
+				Name: name,
+			},
+		},
+	}
+
+	if name == "" {
+		glog.V(100).Info("The name of the Console is empty")
+
+		builder.errorMsg = "errorMsg 'name' cannot be empty"
```
kononovn, line 44: Please change errorMsg to Console.

pkg/console/console.go
```diff
@@ -0,0 +1,176 @@
...
+			ObjectMeta: metaV1.ObjectMeta{
+				Name: name,
+			},
+		},
+	}
+
+	if name == "" {
+		glog.V(100).Info("The name of the Console is empty")
+
+		builder.errorMsg = "'name' cannot be empty"
```
kononovn, line 66: Please add **Console**

pkg/console/console.go
kononovn: Please move this log before > if !builder.Exists()

elenagerman: it was already done

kononovn (approved):
/lgtm

PR #143, opened 2023-09-25
clusterlogging pkg added

elenagerman:


kononovn (changes requested):
@elenagerman Please re-base this PR on top of latest main and we can start review.

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,190 @@
...
+func (builder *ClusterLoggingBuilder) Pull() (*clov1.ClusterLogging, error) {
+	if valid, err := builder.validate(); !valid {
+		return nil, err
+	}
+
+	glog.V(100).Infof("Getting clusterLogging %s in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	clusterLogging := &clov1.ClusterLogging{}
+	err := builder.apiClient.Get(context.Background(), goclient.ObjectKey{
```
kononovn, line 69: Please replace it by following logic:

```
if !builder.Exists() {
	return nil, fmt.Errorf("ClusterLogging object %s doesn't exist", name)
}
```

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,190 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return clusterLogging, err
+}
+
+// Discover fetches existing clusterLogging from cluster.
+func (builder *ClusterLoggingBuilder) Discover() error {
```
kononovn, line 82: Please remove this function.

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	clov1 "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ClusterLoggingBuilder provides struct for clusterLogging object.
+type ClusterLoggingBuilder struct {
```
kononovn, line 17: This is the only 1 object under clusterlogging package. 
Please re-name it to Builder.

pkg/clusterlogging/clusterlogging.go
kononovn: Pleas check Pull functions and use it as example.

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,190 @@
...
+func (builder *ClusterLoggingBuilder) Exists() bool {
+	if valid, _ := builder.validate(); !valid {
+		return false
+	}
+
+	glog.V(100).Infof("Checking if clusterLogging %s exists in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	var err error
+	builder.Object, err = builder.Pull()
```
kononovn, line 155: Use builder.Get() function here

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,190 @@
...
+	if !builder.Exists() {
+		err = builder.apiClient.Create(context.TODO(), builder.Definition)
+		if err == nil {
+			builder.Object = builder.Definition
+		}
+	}
+
+	return builder, err
+}
+
```
kononovn, line 120: Please add Get function to this package

kononovn (changes requested):
Few comments.

pkg/clusterlogging/clusterlogging.go
elenagerman: but this is a Pull function. How you whant me to replace Pull with Exists?
It will be a dead-loop, Exists method uses Pull already

pkg/clusterlogging/clusterlogging.go
elenagerman: I don't wont to add methods I could not test, I add only verified code, I'm sorry.
Just because of the structure you decided each pkg should be build doesn't mean each pkg commit has to include all of it, this is the wrong approach. 
1. Code have to be tested before we commit it
2. not all of the methods, defined as must by your instructions, needed in case of each specific method implementation. Moreover, the implementation of some in specific areas makes the person who writes the code an idiot. Well, like delete clusterversion as an example (I could find more).
Not all such straight forward ...

pkg/clusterlogging/clusterlogging.go
elenagerman: in you presentation two weeks ago it was Pull as a must in a list of methods. Do we start it again?
what is a final agreement with you and Trey: Pull or Get? or both, or maybe Discovery and Get?? or any other combination? I don't really follow your last decisions made...

pkg/clusterlogging/clusterlogging.go
elenagerman: I renamed it to Get

usage/namespace/namespace.go
```diff
@@ -33,11 +33,11 @@ func deleteExistingExample() {
 	if apiClients == nil {
 		panic("Failed to load api client")
 	}
-	// Pull existing namespace from cluster
+	// Get existing namespace from cluster
```
kononovn, line 36: Please remove these changes from commit.

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,170 @@
...
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the clusterLogging is empty")
+
+		builder.errorMsg = "The clusterLogging 'namespace' cannot be empty"
+	}
+
+	return builder
+}
+
```
kononovn, line 58: Don't you need Pull function which will pull existing object to Builder? Refer to example: https://github.com/openshift-kni/eco-goinfra/blob/d37bdd48d168b4737f647c4144bfff5aced29c17/pkg/pod/pod.go#L92C2-L92C2

kononovn (changes requested):
2 more comments.

usage/namespace/namespace.go
elenagerman: ups

pkg/clusterlogging/clusterlogging.go
elenagerman: I have Get instead, it's just the same

kononovn: No get is the different. In order to use get you need to create Builder **first** and then request Get  however Pull will create builder and load object automatically.

kononovn (changes requested):
namespace.go still in commit

elenagerman:
> namespace.go still in commit

it is not 'still', it's just back when I did refactoring for the Pull method name. fixed

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,206 @@
...
+	glog.V(100).Infof("Checking if clusterLogging %s exists in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	var err error
+	builder.Object, err = builder.Get()
+
+	return err == nil || !k8serrors.IsNotFound(err)
+}
+
```
kononovn, line 175: Please update the last function Update.

kononovn (changes requested):
1 last comment.

pkg/clusterlogging/clusterlogging.go
elenagerman: Update what?

kononovn (dismissed):
lgtm

pkg/clusterlogging/clusterlogging.go
```diff
@@ -0,0 +1,240 @@
...
+					"Failed to update the clusterLogging object %s in namespace $s."+
+						"due to error in delete function", builder.Definition.Name, builder.Definition.Namespace)
+
+				return nil, err
+			}
+
+			return builder.Create()
+		}
+	}
+
```
trewest, line 206:
```suggestion
if err == nil {
    builder.Object = builder.Definition
}
```

trewest (changes requested):
Looks good, just one comment

trewest (approved):
/lgtm

kononovn (approved):
lgtm

PR #150, opened 2023-09-27
assisted pkg: changes to Update functions

ElaiShalevRH:
Added an assignment to builder.Object in the infraenv, agentclusterinstall, agent and agentserviceconfig files. Added a Delete and Wait function for the infraenv.

### DIFF
--- a/pkg/assisted/agent.go
+++ b/pkg/assisted/agent.go
@@ -308,6 +308,8 @@ func (builder *agentBuilder) Update() (*agentBuilder, error) {
 		builder.Object = builder.Definition
 	}
 
+	builder.Object = builder.Definition
+
 	return builder, err
 }
 
--- a/pkg/assisted/agent.go
+++ b/pkg/assisted/agent.go
@@ -308,8 +308,6 @@ func (builder *agentBuilder) Update() (*agentBuilder, error) {
 		builder.Object = builder.Definition
 	}
 
-	builder.Object = builder.Definition
-
 	return builder, err
 }
 
--- a/pkg/assisted/agentserviceconfig.go
+++ b/pkg/assisted/agentserviceconfig.go
@@ -381,6 +381,8 @@ func (builder *AgentServiceConfigBuilder) Update(force bool) (*AgentServiceConfi
 		}
 	}
 
+	builder.Object = builder.Definition
+
 	return builder, err
 }
 
--- a/pkg/assisted/agentserviceconfig.go
+++ b/pkg/assisted/agentserviceconfig.go
@@ -381,7 +381,9 @@ func (builder *AgentServiceConfigBuilder) Update(force bool) (*AgentServiceConfi
 		}
 	}
 
-	builder.Object = builder.Definition
+	if err == nil {
+		builder.Object = builder.Definition
+	}
 
 	return builder, err
 }
--- a/pkg/assisted/infraenv.go
+++ b/pkg/assisted/infraenv.go
@@ -783,7 +783,7 @@ func (builder *InfraEnvBuilder) Update(force bool) (*InfraEnvBuilder, error) {
 				builder.Definition.Name, builder.Definition.Namespace,
 			)
 
-			_, err = builder.Delete()
+			err = builder.DeleteAndWait(time.Second * 5)
 			builder.Definition.ResourceVersion = ""
 
 			if err != nil {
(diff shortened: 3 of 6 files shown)


pkg/assisted/agent.go
```diff
@@ -308,6 +308,8 @@ func (builder *agentBuilder) Update() (*agentBuilder, error) {
 		builder.Object = builder.Definition
 	}
 
+	builder.Object = builder.Definition
```
trewest, line 311: Seem like this is already being done on #L308. You can remove it from here

pkg/assisted/agentserviceconfig.go
```diff
@@ -381,6 +381,8 @@ func (builder *AgentServiceConfigBuilder) Update(force bool) (*AgentServiceConfi
 		}
 	}
 
+	builder.Object = builder.Definition
```
trewest, line 384:
```suggestion
	if err == nil {
		builder.Object = builder.Definition
	}
```

pkg/assisted/infraenv.go
```diff
@@ -800,6 +800,8 @@ func (builder *InfraEnvBuilder) Update(force bool) (*InfraEnvBuilder, error) {
 		}
 	}
 
+	builder.Object = builder.Definition
```
trewest, line 803:
```suggestion
		if err == nil {
		builder.Object = builder.Definition
	}
```

trewest (changes requested):
Just a few minor comments

pkg/assisted/infraenv.go
```diff
@@ -827,6 +831,32 @@ func (builder *InfraEnvBuilder) Delete() (*InfraEnvBuilder, error) {
...
+		builder.Definition.Name)
+
+	if _, err := builder.Delete(); err != nil {
+		return err
+	}
+
+	// Polls the InfraEnvBuilder every second until it's removed.
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		_, err := builder.Get()
+		if k8serrors.IsNotFound(err) {
```
achuzhoy, line 851: IsNotFound returns true if the specified error was created by NewNotFound.
I don't find the use of NewNotFound

ElaiShalevRH: Got you, I'll change it to a generic error handling, thanks

ElaiShalevRH:
> @ElaiShalevRH Looks good. Can you also update the hive clusterdeployment and clusterimageset to match?

trewest (dismissed):
/lgtm

pkg/assisted/infraenv.go
```diff
@@ -827,6 +831,32 @@ func (builder *InfraEnvBuilder) Delete() (*InfraEnvBuilder, error) {
 	return builder, nil
 }
 
+// DeleteAndWait deletes an InfraEnvBuilder and waits until it is removed from the cluster.
+func (builder *InfraEnvBuilder) DeleteAndWait(timeout time.Duration) error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof(`Deleting InfraEnvBuilder %s and
```
achuzhoy, line 840: Deleting Infraenv

pkg/assisted/infraenv.go
```diff
@@ -827,6 +831,32 @@ func (builder *InfraEnvBuilder) Delete() (*InfraEnvBuilder, error) {
...
+	if _, err := builder.Delete(); err != nil {
+		return err
+	}
+
+	// Polls the InfraEnvBuilder every second until it's removed.
+	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
+		_, err := builder.Get()
+		if err != nil {
+
+			return true, nil
```
achuzhoy, line 853: shouldn't this be "false, err"

ElaiShalevRH: Yes, correct. I'll change it thanks

trewest: This should be true. If err != nil, that means an error occurred because the InfraEnv doesn't exist which is what we want in this case

achuzhoy: right.

achuzhoy (approved):

trewest (approved):
/lgtm

PR #151, opened 2023-09-27
networkpolicy pkg: fix WithPolicyType method

elenagerman:
fix WithPolicyType method: setting up type value directly to the Object instead of creating an empty map when policyTypes list is empty

### DIFF
--- a/pkg/networkpolicy/networkpolicy.go
+++ b/pkg/networkpolicy/networkpolicy.go
@@ -131,7 +131,9 @@ func (builder *NetworkPolicyBuilder) WithPolicyType(policyType netv1.PolicyType)
 	}
 
 	if builder.Definition.Spec.PolicyTypes == nil {
-		builder.Definition.Spec.PolicyTypes = make([]netv1.PolicyType, 1)
+		builder.Definition.Spec.PolicyTypes = []netv1.PolicyType{policyType}
+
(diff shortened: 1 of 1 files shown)


pkg/networkpolicy/networkpolicy.go
```diff
@@ -131,11 +131,11 @@ func (builder *NetworkPolicyBuilder) WithPolicyType(policyType netv1.PolicyType)
 	}
 
 	if builder.Definition.Spec.PolicyTypes == nil {
-		builder.Definition.Spec.PolicyTypes = make([]netv1.PolicyType, 1)
+		builder.Definition.Spec.PolicyTypes = []netv1.PolicyType{policyType}
+	} else {
```
kononovn, line 135: Please remove else and add return
```
if builder.Definition.Spec.PolicyTypes == nil {
    builder.Definition.Spec.PolicyTypes = []netv1.PolicyType{policyType}
    retrun builder
}

builder.Definition.Spec.PolicyTypes = append(builder.Definition.Spec.PolicyTypes, policyType)

return builder
```

elenagerman: I have no idea why you prefer it like this

kononovn (approved):
lgtm

PR #154, opened 2023-10-04
pod pkg: added GetFullLog function

ElaiShalevRH:
The function allows to get a pod container's full log without passing a time-frame

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -952,6 +952,33 @@ func (builder *Builder) GetLog(logStartTime time.Duration, containerName string)
 	return buf.String(), nil
 }
 
+// GetFullLog connects to a pod and fetches the full log since pod creation.
+func (builder *Builder) GetFullLog(containerName string) (string, error) {
+	if valid, err := builder.validate(); !valid {
+		return "", err
+	}
+
+	req := builder.apiClient.Pods(builder.Definition.Namespace).GetLogs(builder.Definition.Name, &v1.PodLogOptions{})
+	log, err := req.Stream(context.Background())
+
+	if err != nil {
+		return "", err
+	}
+
+	defer func() {
+		_ = log.Close()
+	}()
+
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, log)
+
+	if err != nil {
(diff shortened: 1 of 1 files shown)


ElaiShalevRH:
@trewest @achuzhoy

pkg/pod/pod.go
```diff
@@ -952,6 +952,33 @@ func (builder *Builder) GetLog(logStartTime time.Duration, containerName string)
 	return buf.String(), nil
 }
 
+// GetFullLog connects to a pod and fetches the full log since pod creation.
+func (builder *Builder) GetFullLog(containerName string) (string, error) {
+	if valid, err := builder.validate(); !valid {
+		return "", err
+	}
+
+	req := builder.apiClient.Pods(builder.Definition.Namespace).GetLogs(builder.Definition.Name, &v1.PodLogOptions{})
```
achuzhoy, line 961: can we get a more meaningful var name here?

pkg/pod/pod.go
```diff
@@ -952,6 +952,33 @@ func (builder *Builder) GetLog(logStartTime time.Duration, containerName string)
...
+
+	if err != nil {
+		return "", err
+	}
+
+	defer func() {
+		_ = log.Close()
+	}()
+
+	buf := new(bytes.Buffer)
```
achuzhoy, line 972: more meaning var name would help future readers.

pkg/pod/pod.go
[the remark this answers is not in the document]
ElaiShalevRH: Thanks, you're right, I'll fix it

pkg/pod/pod.go
ElaiShalevRH: @trewest @achuzhoy Honestly I just copied the GetLog function and dropped the 'SinceSeconds' field. I didn't touch any variable names from what was already on pod pkg
I can of course change the variable names but maybe we should do it as well for the first function?

achuzhoy: IMHO The other function can (or even should) be a separate PR.

trewest (approved):
/lgtm

PR #155, opened 2023-10-09
Add catalog info for backstage

cdvultur:
Adding catalog infor for backstage.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,23 @@
+# Catalog entry for Backstage [backstage.io]
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: eco-goinfra
+  title: eco-goinfra
+  description: Ecosystem QE infrastructure framework
+  links:
+    - title : Golang
+      url: https://go.dev/
+  tags:
+    - golang
+    - test-framework
+    - product-quality
+    - openshift
+  namespace: quality-community
+  annotations:
+    github.com/project-slug: openshift-kni/eco-goinfra
+    backstage.io/techdocs-ref: https://github.com/openshift-kni/eco-goinfra/blob/main/README.md
+spec:
+  type: library
+  owner: group:redhat/openshift-kni
+  lifecycle: production


catalog-info.yaml
```diff
@@ -0,0 +1,24 @@
...
+  tags:
+    - go
+    - ginkgo
+    - test-framework
+    - product-quality
+    - openshift
+  namespace: quality-community
+  annotations:
+    github.com/project-slug: openshift-kni/eco-goinfra
+    backstage.io/techdocs-ref: https://github.com/openshift-kni/eco-goinfra/wiki
```
trewest, line 20: Should we have something here?

cdvultur: From what I see in other examples we should point to the documentation. 

Maybe it is better to have it point to the README ?

catalog-info.yaml
```diff
@@ -0,0 +1,24 @@
+# Catalog entry for Backstage [backstage.io]
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: eco-goinfra
+  title: eco-goinfra
+  description: Ecosystem QE framework
+  links:
+    - title : Ginkgo
```
kononovn, line 9: This set of tools doesn't use ginkgo

catalog-info.yaml
trewest: Yeah I think the readme makes sense

cdvultur: Updated to use README.

catalog-info.yaml
cdvultur: Correct! Updated it to go.dev, so that it is in sync with Readme

trewest (approved):
/lgtm

kononovn (changes requested):
1 comment

cdvultur:
Updated. Thanks !

catalog-info.yaml
```diff
@@ -0,0 +1,23 @@
+# Catalog entry for Backstage [backstage.io]
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: eco-goinfra
+  title: eco-goinfra
+  description: Ecosystem QE framework
```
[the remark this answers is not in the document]
cdvultur: Updated it.

Thanks

kononovn (approved):
/lgtm

PR #156, opened 2023-10-09
Add StorageClassBuilder to storage pkg

trewest:
Adds the StorageClassBuilder struct to the storage pkg

### DIFF
--- a/pkg/storage/storageclass.go
+++ b/pkg/storage/storageclass.go
@@ -1,0 +1,301 @@
+package storage
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	v1 "k8s.io/api/core/v1"
+	storageV1 "k8s.io/api/storage/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClassBuilder provides struct for storageclass object containing
+// connection to the cluster and the storageclass definitions.
+type ClassBuilder struct {
+	// Storageclass definition. Used to create the storageclass object.
+	Definition *storageV1.StorageClass
+	// Created storageclass object
+	Object *storageV1.StorageClass
+	// Used in functions that define or mutate storageclass definition. errorMsg is processed before the storageclass
+	// object is created.
+	errorMsg  string
+	apiClient *clients.Settings
+}
+
+// AdditionalOptions additional options for storageclass object.
+type AdditionalOptions func(builder *ClassBuilder) (*ClassBuilder, error)
+
+// NewClassBuilder creates a new instance of ClassBuilder.
+func NewClassBuilder(apiClient *clients.Settings, name, provisioner string) *ClassBuilder {
+	glog.V(100).Infof(
+		"Initializing new storageclass structure with the following params: "+
+			"name: %s, provisioner: %s", name, provisioner)
+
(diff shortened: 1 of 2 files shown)


pkg/storage/storageclass.go
```diff
@@ -0,0 +1,266 @@
...
+	// object is created.
+	errorMsg  string
+	apiClient *clients.Settings
+}
+
+// AdditionalOptions additional options for storageclass object.
+type AdditionalOptions func(builder *StorageClassBuilder) (*StorageClassBuilder, error)
+
+// NewStorageClassBuilder creates a new instance of StorageClassBuilder.
+func NewStorageClassBuilder(apiClient *clients.Settings, name, provisioner string) *StorageClassBuilder {
```
kononovn, line 33: I think it's name duplication. You can simply re-name it to NewClassBuilder because when you call this function you have storage in package name.
`storageClass := storage.NewClassBuilder`

pkg/storage/storageclass.go
```diff
@@ -0,0 +1,266 @@
...
+	return builder, err
+}
+
+// Delete removes a storageclass.
+func (builder *StorageClassBuilder) Delete() error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof("Deleting storageclass %s",
```
kononovn, line 202: Could be single line.

pkg/storage/storageclass.go
```diff
@@ -0,0 +1,266 @@
...
+		return err
+	}
+
+	builder.Object = nil
+
+	return err
+}
+
+// Update renovates the existing storageclass object with the storageclass definition in builder.
+func (builder *StorageClassBuilder) Update() (*StorageClassBuilder, error) {
```
kononovn, line 222: Have you managed to test this function? Does it work ?

kononovn (changes requested):
Overall looks good, just few minor comments

cdvultur (approved):
/lgtm

pkg/storage/storageclass.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+
+	if name == "" {
+		glog.V(100).Infof("The name of the storageclass is empty")
+
+		builder.errorMsg = "storageclass 'name' cannot be empty"
+	}
+
+	if provisioner == "" {
+		glog.V(100).Infof("The provisioner of the storageclass is empty")
```
achuzhoy, line 55: maybe set it yourself to "kubernetes.io/no-provisioner" instead.

trewest: Not sure I follow. kubernetes.io/no-provisioner is only one possible provisioner

pkg/storage/storageclass.go
```diff
@@ -0,0 +1,301 @@
...
+}
+
+// WithParameter adds a parameter to the storage class definition.
+func (builder *ClassBuilder) WithParameter(parameterKey, parameterValue string) *ClassBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	if parameterKey == "" {
+		glog.V(100).Infof("The parameterKey of the storageclass is empty")
```
achuzhoy, line 114: nit: there's no notion of parameterKey nor parameterValue for sc.
I'd rephrase to:
the "parameter's key" and "parameter's value" respectively

pkg/storage/storageclass.go
achuzhoy: "no-provisioner" I'd consider using it as default maybe.

trewest: I personally think it should just be provided by the user. Otherwise the user needs some previous knowledge of what we have chosen the default to be which could be confusing.

achuzhoy (approved):

PR #158, opened 2023-10-12
Add GetEvents method for AgentClusterInstall

trewest:
Adds a GetEvents method for AgentClusterInstall. Will be used for determining state of installed/installing clusters


cdvultur (approved):
/lgtm

PR #159, opened 2023-10-13
initialize kmm args/rawargs only when length is greater than 0

cdvultur:
Calling WithModprobeSpec will set empty args/rawArgs, causing module loading problem.


qbarrand (approved):

trewest (approved):
/lgtm

kononovn (approved):
/lgtm

PR #162, opened 2023-10-16
infrastructure pkg added

elenagerman:
infrastructure pkg added

### DIFF
--- a/pkg/infrastructure/infrastructure.go
+++ b/pkg/infrastructure/infrastructure.go
@@ -1,0 +1,90 @@
+package infrastructure
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	v1 "github.com/openshift/api/config/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	infrastructureName = "cluster"
+)
+
+// Builder provides a struct for infrastructure object from the cluster and a infrastructure definition.
+type Builder struct {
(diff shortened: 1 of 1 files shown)


pkg/infrastructure/infrastructure.go
```diff
@@ -0,0 +1,92 @@
...
+	return &builder, nil
+}
+
+// Exists checks whether the given infrastructure exists.
+func (builder *Builder) Exists() bool {
+	if valid, _ := builder.validate(); !valid {
+		return false
+	}
+
+	glog.V(100).Infof(
```
cdvultur, line 57: To have it consistent with the rest of `Infof()`calls, it would be nice to have this in one line. 

Seems to be the only one diverting from the pattern.

pkg/infrastructure/infrastructure.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Builder provides a struct for infrastructure object from the cluster and a infrastructure definition.
+type Builder struct {
+	// infrastructure definition, used to create the infrastructure object.
+	Definition *v1.Infrastructure
+	// Created infrastructure object.
+	Object *v1.Infrastructure
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
+}
```
cdvultur, line 27: Don't we want to have errorMsg ?

cdvultur (commented):
Wondering if Update() function might not be useful to add.

pkg/infrastructure/infrastructure.go
elenagerman: I don't see where we could use it in this case

elenagerman:
> Wondering if Update() function might not be useful to add.

we are not creating or modifying this object, imho it could be usefull

pkg/infrastructure/infrastructure.go
cdvultur: Currently it seems not needed. Still was related with the `Update()` discussion.

cdvultur (approved):
/lgtm

PR #164, opened 2023-10-17
clusterloggingforwarding package added to the clusterlogging

elenagerman:
clusterloggingforwarding package added to the clusterlogging

### DIFF
--- a/pkg/clusterlogging/clusterlogforwarder.go
+++ b/pkg/clusterlogging/clusterlogforwarder.go
@@ -1,0 +1,205 @@
+package clusterlogging
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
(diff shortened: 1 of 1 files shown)


pkg/olm/clusterserviceversion.go
```diff
@@ -3,6 +3,7 @@ package olm
 import (
 	"context"
 	"fmt"
+
```
trewest, line 6: Accident?

trewest (changes requested):
Just one nit, otherwise looks good

pkg/olm/clusterserviceversion.go
elenagerman: no, there is no explanation, but lint verification fails on fmt for clusterserviceversion, so I have to fix it each time

trewest (approved):
/lgtm

kononovn (approved):
/lgtm

PR #165, opened 2023-10-18
add WithVolume method to the deployment pkg

elenagerman:
add WithVolumes method to the deployment pkg

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -76,12 +76,6 @@ func NewBuilder(
 		builder.errorMsg = "deployment 'namespace' cannot be empty"
 	}
 
-	if len(labels) == 0 {
-		glog.V(100).Infof("There are no labels for the deployment")
-
-		builder.errorMsg = "deployment 'labels' cannot be empty"
-	}
-
 	return &builder
 }
 
@@ -315,6 +309,54 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
 	return builder
 }
 
+// WithVolume sets the Volume on the deployment definition.
+func (builder *Builder) WithVolume(volume *coreV1.Volume) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Configuring volume %v on deployment %s in namespace %s",
+		volume, builder.Definition.Name, builder.Definition.Namespace)
+
+	if volume == nil {
+		glog.V(100).Infof("The 'volume' of the deployment is empty")
+
+		builder.errorMsg = "'volume' parameter is empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	mountConfig := coreV1.VolumeMount{Name: volume.Name, MountPath: volume.HostPath.Path, ReadOnly: false}
+
+	builder.isMountAlreadyInUseInDeployment(mountConfig)
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	for index := range builder.Definition.Spec.Template.Spec.Containers {
+		builder.Definition.Spec.Template.Spec.Containers[index].VolumeMounts = append(
+			builder.Definition.Spec.Template.Spec.Containers[index].VolumeMounts, mountConfig)
+	}
+
+	if len(builder.Definition.Spec.Template.Spec.InitContainers) > 0 {
+		for index := range builder.Definition.Spec.Template.Spec.InitContainers {
+			builder.Definition.Spec.Template.Spec.InitContainers[index].VolumeMounts = append(
+				builder.Definition.Spec.Template.Spec.InitContainers[index].VolumeMounts, mountConfig)
+		}
+	}
+
+	if builder.Definition.Spec.Template.Spec.Volumes == nil {
+		builder.Definition.Spec.Template.Spec.Volumes = []coreV1.Volume{*volume}
+	} else {
+		builder.Definition.Spec.Template.Spec.Volumes = append(builder.Definition.Spec.Template.Spec.Volumes, *volume)
+	}
+
+	return builder
+}
+
 // WithOptions creates deployment with generic mutation options.
 func (builder *Builder) WithOptions(options ...AdditionalOptions) *Builder {
 	if valid, _ := builder.validate(); !valid {
@@ -529,6 +571,29 @@ func (builder *Builder) WaitUntilCondition(condition v1.DeploymentConditionType,
 // GetGVR returns deployment's GroupVersionResource which could be used for Clean function.
 func GetGVR() schema.GroupVersionResource {
 	return schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"}
+}
+
+func (builder *Builder) isMountAlreadyInUseInDeployment(newMount coreV1.VolumeMount) {
+	if valid, _ := builder.validate(); valid {
+		for index := range builder.Definition.Spec.Template.Spec.Containers {
+			if builder.Definition.Spec.Template.Spec.Containers[index].VolumeMounts != nil {
+				if isMountInUse(builder.Definition.Spec.Template.Spec.Containers[index].VolumeMounts, newMount) {
+					builder.errorMsg = fmt.Sprintf("given mount %v already mounted to pod's container %s",
+						newMount.Name, builder.Definition.Spec.Template.Spec.Containers[index].Name)
+				}
+			}
+		}
+	}
+}
+
+func isMountInUse(containerMounts []coreV1.VolumeMount, newMount coreV1.VolumeMount) bool {
+	for _, containerMount := range containerMounts {
+		if containerMount.Name == newMount.Name && containerMount.MountPath == newMount.MountPath {
+			return true
+		}
+	}
+
+	return false
 }
 
 // validate will check that the builder and builder definition are properly initialized before


pkg/deployment/deployment.go
```diff
@@ -315,6 +315,34 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
 	return builder
 }
 
+// WithVolumes sets the Volume on the deployment definition.
+func (builder *Builder) WithVolumes(volume *coreV1.Volume) *Builder {
```
kononovn, line 319: There is only 1 volume in argument. Please re-name the function to **WithVolume**

kononovn (changes requested):
Just 1 comment  to fix.

pkg/deployment/deployment.go
elenagerman: the final parameter name is volumes, but as you wish

kononovn (approved):
/lgtm

pkg/deployment/deployment.go
```diff
@@ -315,6 +315,34 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
 	return builder
 }
 
+// WithVolume sets the Volume on the deployment definition.
+func (builder *Builder) WithVolume(volume *coreV1.Volume) *Builder {
```
trewest, line 319: I see that this code adds a volume to a deployment but I don't see how any containers within the deployment can consume it. I think we should include another method that accepts a container name and a *coreV1.VolumeMount in order to make use of the volume. WDYT?

kononovn: @trewest Do you suggest something like that? >
https://github.com/openshift-kni/eco-goinfra/blob/8a2c85bc508cd911d35c4ac203df41ea6330b46b/pkg/pod/pod.go#L603C5-L603C5

trewest: Yeah, the only thing is that this method is limited to configmap volume sources so it may get a bit difficult to write a method for each volume source type and all of their supported parameters (there are many variations)

kononovn: Can you please share what you propose?

elenagerman: it is not so clear how you think we could how can we relate one to the other?
WE already have the method dedicated to the container configuration: WithAdditionalContainerSpecs.
What you ask me to do will add a complicity and I believe end-user has to take care about it.

elenagerman: and, actually, I have no idea how we could implement such verification when two different methods involved in obj creation and object is not created yet

trewest: @kononovn First I think we need to answer the question of do we only support mapping a volume to all containers in a pod or should the user be able to choose which containers the volume is attached to? For instance, two containers may be able to use the same volume but the mountPath in each of those containers could be different, this means we would have to decouple the WithVolume and WithVolumeMount operations. Take for instance:
```
	deployment.NewBuilder(APIClient, "test-deployment", "test-deployment", map[string]string{"app": "test-deployment"}, &v1.Container{
		Name:  "web",
		Image: "nginx",
	}).WithAdditionalContainerSpecs([]v1.Container{
		{
			Name:  "database",
			Image: "postgres",
		},
	})

	userVol := &v1.Volume{
		Name: "uservol",
		VolumeSource: v1.VolumeSource{
			ConfigMap: &v1.ConfigMapVolumeSource{
				LocalObjectReference: v1.LocalObjectReference{
					Name: "user-data-configmap",
				},
			},
		},
	}

	deployment.WithVolume(userVol).WithVolumeMount("web", userVol.Name, v1.VolumeMount{
		ReadOnly:  true,
		MountPath: "/usr/share/nginx/users",
	}).WithVolumeMount("database", userVol.Name, v1.VolumeMount{
		ReadOnly:  true,
		MountPath: "/etc/postgresql/users.conf",
	}).Create()
```

So if we wanted to support something like this, the methods should look like
```
func (builder *Builder) WithVolume(volume *coreV1.Volume) *Builder {}
func (builder *Builder) WithVolumeMount(containerName, volumeName string, volumeMount *coreV1.VolumeMount) *Builder {}
```

kononovn: @trewest Let's support  mapping a volume to all containers  for now and if later we got requirements or real use-case scenario we can change it by adding new API method. Something like WithVolumeInContainer or WithVolumeConfig. WDYT ?

trewest: Okay if that is the case, let's do it like the pod [pkg](https://github.com/openshift-kni/eco-goinfra/blob/8a2c85bc508cd911d35c4ac203df41ea6330b46b/pkg/pod/pod.go#L603C5-L603C5) but take in a volume rather than only supporting configmap. Sound good?
```
func (builder *Builder) WithVolume(volume *coreV1.Volume, mountPath string ) *Builder {}
```
We should just verify that `volume != nil && volume.Name != ""`

kononovn: @trewest  Ok agree +1. 
@elenagerman Please refactor this function according to this example https://github.com/openshift-kni/eco-goinfra/blob/8a2c85bc508cd911d35c4ac203df41ea6330b46b/pkg/pod/pod.go#L603C5-L603C5 . Please note that you don't need to change current argument **volume *coreV1.Volume** to **volumeName**
The main idea of refactoring is to make sure that the volume is forwarded in to all containers under deployment.

pkg/deployment/deployment.go
```diff
@@ -315,6 +309,54 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
...
```
trewest, line 313:
```suggestion
func (builder *Builder) WithVolume(volume *coreV1.Volume, mountPath string) *Builder {
```

pkg/deployment/deployment.go
```diff
@@ -315,6 +309,54 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
...
+		glog.V(100).Infof("The 'volume' of the deployment is empty")
+
+		builder.errorMsg = "'volume' parameter is empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	mountConfig := coreV1.VolumeMount{Name: volume.Name, MountPath: volume.HostPath.Path, ReadOnly: false}
```
trewest, line 331:
```suggestion
	mountConfig := coreV1.VolumeMount{Name: volume.Name, MountPath: mountPath, ReadOnly: false}
```

pkg/deployment/deployment.go
```diff
@@ -315,6 +309,54 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
...
+
+	glog.V(100).Infof("Configuring volume %v on deployment %s in namespace %s",
+		volume, builder.Definition.Name, builder.Definition.Namespace)
+
+	if volume == nil {
+		glog.V(100).Infof("The 'volume' of the deployment is empty")
+
+		builder.errorMsg = "'volume' parameter is empty"
+	}
+
```
trewest, line 326: Please also check that volume.Name != ""

trewest (changes requested):
Just a couple of comments

pkg/deployment/deployment.go
elenagerman: we don't need mount path, it is already inside volume variable:
```go
MountPath: volume.HostPath.Path
```

pkg/deployment/deployment.go
elenagerman: I don't understand why?

elenagerman: you told me do not change type of volume to the volumeName. If it stay to be volume *coreV1.Volume object, so we have all inside

elenagerman: volume.HostPath.Path is a required value

elenagerman:
please be consistent in your comments. you ask me do not expand such big types of objects. In this specific
```go
coreV1.Volume{
		Name:         "",
		VolumeSource: coreV1.VolumeSource{
			HostPath:              nil,
			EmptyDir:              nil,
			GCEPersistentDisk:     nil,
			AWSElasticBlockStore:  nil,
			GitRepo:               nil,
			Secret:                nil,
			NFS:                   nil,
			ISCSI:                 nil,
			Glusterfs:             nil,
			PersistentVolumeClaim: nil,
			RBD:                   nil,
			FlexVolume:            nil,
			Cinder:                nil,
			CephFS:                nil,
			Flocker:               nil,
			DownwardAPI:           nil,
			FC:                    nil,
			AzureFile:             nil,
			ConfigMap:             nil,
			VsphereVolume:         nil,
			Quobyte:               nil,
			AzureDisk:             nil,
			PhotonPersistentDisk:  nil,
			Projected:             nil,
			PortworxVolume:        nil,
			ScaleIO:               nil,
			StorageOS:             nil,
			CSI:                   nil,
			Ephemeral:             nil,
		},
	}
```  case:

File: pkg/deployment/deployment.go (hunk 3 above)
Comment (trewest) [reply pkg/deployment/deployment.go:313]: These values aren't equivalent:
https://pkg.go.dev/k8s.io/api/core/v1#VolumeMount
https://pkg.go.dev/k8s.io/api/core/v1#Volume
By using volume.HostPath.Path as the mountPath you are effectively limiting this method to only working with HostPath volumes.

File: pkg/deployment/deployment.go (hunk 4 above)
Comment (trewest) [reply pkg/deployment/deployment.go:331]: Refer to my comment here https://github.com/openshift-kni/eco-goinfra/pull/165#discussion_r1385386274

**Comment by trewest** [comment]
> please be consistent in your comments. you ask me do not expand such big types of objects. In this specific

Not sure I follow

File: pkg/deployment/deployment.go (hunk 4 above)
Comment (trewest) [reply pkg/deployment/deployment.go:331]: I mentioned the expected signature for the method here https://github.com/openshift-kni/eco-goinfra/pull/165#discussion_r1367327935

Comment (trewest) [reply pkg/deployment/deployment.go:331]: > volume.HostPath.Path is a required value

Not sure where you're seeing that it is a required value for [VolumeSource](https://pkg.go.dev/k8s.io/api/core/v1#VolumeSource) resources. Can you share?

**Comment by elenagerman** [comment]
https://github.com/openshift-kni/eco-goinfra/pull/147#discussion_r1362659894

**Comment by elenagerman** [comment]
@trewest volume object of coreV1.Volume type includes all needed information for creating volume

**Comment by trewest** [comment]
> [#147 (comment)](https://github.com/openshift-kni/eco-goinfra/pull/147#discussion_r1362659894)

What I am saying here is consistent with this comment. In this scenario *coreV1.Volume is the complex struct and instead of also requiring a *coreV1.VolumeMount parameter, we are simplifying it to a single string parameter (mountPath) and setting the name based on the volume parameter and hardcoding ReadOnly to false:
```
coreV1.VolumeMount{Name: volume.Name, MountPath: mountPath, ReadOnly: false}
```

**Comment by trewest** [comment]
> @trewest volume object of coreV1.Volume type includes all needed information for creating volume

I'm not sure how. If I wanted to use a volume like my example above:
```
&v1.Volume{
		Name: "uservol",
		VolumeSource: v1.VolumeSource{
			ConfigMap: &v1.ConfigMapVolumeSource{
				LocalObjectReference: v1.LocalObjectReference{
					Name: "user-data-configmap",
				},
			},
		},
	}
```
The WithVolume method as currently written would not work

elenagerman:
@trewest 
The method is called WithVolume, not WithVolumeMount. The meaning is:  ALL available variations of Volume object configuration could be used. We could not limit it to the mount will have to be provided

pkg/deployment/deployment.go
elenagerman: Error: pkg/deployment/deployment.go:327:12: SA5011: possible nil pointer dereference (staticcheck)
level=info msg="[runner] processing took 11.791277ms with stages: exclude-rules: 3.062368ms, path_prettifier: 2.369276ms, nolint: 2.271976ms, identifier_marker: 2.024779ms, autogenerated_exclude: 1.638283ms, skip_dirs: 350.096µs, source_code: 45.3µs, cgo: 11.299µs, filename_unadjuster: 5.1µs, max_same_issues: 3.2µs, uniq_by_line: 2.7µs, max_per_file_from_linter: 1.4µs, path_shortener: 1.2µs, max_from_linter: 1.2µs, fixer: 800ns, skip_files: 600ns, exclude: 400ns, diff: 400ns, severity-rules: 400ns, sort_results: 300ns, path_prefixer: 200ns"
level=info msg="[runner] linters took 1m45.933192091s with stages: goanalysis_metalinter: 1m45.92085012s"
level=info msg="File cache stats: 108 entries of total size 733.9KiB"
	if volume.Name == "" {
	          ^

elenagerman:
@trewest @kononovn I close this commit. With all changes you require it break all possible logic of this method.
I will apply this config via cmd line

PR #175, opened 2023-10-24
clusterrole: remove rules verification for the empty fields

elenagerman:
The rule does not include each available field; some can stay empty. As an example:

```go

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: {{ cr_name }}
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  - nodes/proxy
  - services
  - endpoints
  - pods
  - namespaces
  verbs:
  - get
  - list
  - watch
- nonResourceURLs:
  - /metrics
  - /federate
  verbs:
  - get
```

### DIFF
--- a/pkg/rbac/clusterrole.go
+++ b/pkg/rbac/clusterrole.go
@@ -79,22 +79,10 @@ func (builder *ClusterRoleBuilder) WithRules(rules []v1.PolicyRule) *ClusterRole
 	}
 
 	for _, rule := range rules {
-		if len(rule.APIGroups) == 0 {
-			glog.V(100).Infof("The clusterrole rule must contain at least one APIGroup entry")
-
-			builder.errorMsg = "clusterrole rule must contain at least one APIGroup entry"
-		}
-
 		if len(rule.Verbs) == 0 {
 			glog.V(100).Infof("The clusterrole rule must contain at least one Verb entry")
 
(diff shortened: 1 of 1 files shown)


achuzhoy (dismissed):

pkg/rbac/clusterrole.go
```diff
@@ -78,30 +78,6 @@ func (builder *ClusterRoleBuilder) WithRules(rules []v1.PolicyRule) *ClusterRole
...
 	}
 
-	for _, rule := range rules {
-		if len(rule.APIGroups) == 0 {
-			glog.V(100).Infof("The clusterrole rule must contain at least one APIGroup entry")
-
-			builder.errorMsg = "clusterrole rule must contain at least one APIGroup entry"
-		}
-
-		if len(rule.Verbs) == 0 {
```
trewest, line 88: Seems that Verbs is a required field. I think we should keep it

trewest (changes requested):
One comments

pkg/rbac/clusterrole.go
elenagerman: yes, I think you're right, verbs verification should be left

trewest (approved):
/lgtm

PR #178, opened 2023-11-01
assisted ztp: infraenv check for populated isodownloadurl

achuzhoy:
The validtion function was missing this check.


achuzhoy:
cc @trewest @vkolodny

trewest:
This may be relevant for some situations but it isn't appropriate for the validate method. For example, none of the With methods will work for creating an initial InfraEnv. They would only work for updating an existing InfraEnv

PR #179, opened 2023-11-02
Added Acm package with to support policy objects

shaior:


pkg/acm/policy.go
```diff
@@ -0,0 +1,239 @@
+package acm
```
cdvultur, line 1: I think it would be better to have this called `ocm` because as it seems it uses open-cluster-management.io

pkg/acm/policy.go
```diff
@@ -0,0 +1,239 @@
...
+		glog.V(100).Infof("The %s builder has error message: %s", resourceCRD, builder.errorMsg)
+
+		return false, fmt.Errorf(builder.errorMsg)
+	}
+
+	return true, nil
+}
+
+// ListPoliciesInAllNamespaces returns a cluster-wide pod inventory.
+func ListPoliciesInAllNamespaces(apiClient *clients.Settings, options metaV1.ListOptions) ([]*PolicyBuilder, error) {
```
cdvultur, line 212: I think the List should be part of a different file ( policylist.go ) that will handle lists of policies .

There are few examples in different packages.

cdvultur (changes requested):
Overall looks good, still some changes are needed so that it is coherent with existing packages

cdvultur (approved):
/lgtm

pkg/ocm/list.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ListPoliciesInAllNamespaces returns a cluster-wide policy inventory.
+func ListPoliciesInAllNamespaces(apiClient *clients.Settings, options metaV1.ListOptions) ([]*PolicyBuilder, error) {
```
trewest, line 14: It seems that you are not using the options parameter that is provided

pkg/ocm/list.go
```diff
@@ -0,0 +1,41 @@
+package ocm
```
trewest, line 1: Can you rename this file policylist.go? I'm expecting there to be more structs added to this pkg

pkg/ocm/policy.go
```diff
@@ -0,0 +1,209 @@
...
+					"Failed to update the policy object %s, "+
+						"due to error in delete function", builder.Definition.Name)
+
+				return nil, err
+			}
+
+			return builder.Create()
+		}
+	}
+
```
trewest, line 175:
```suggestion
	if err == nil {
		builder.Object = builder.Definition
	}
```

trewest (changes requested):
Just a few comments

pkg/ocm/list.go
shaior: Correct, thanks.

trewest (approved):
/lgtm

pkg/ocm/policylist.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ListPoliciesInAllNamespaces returns a cluster-wide policy inventory.
+func ListPoliciesInAllNamespaces(apiClient *clients.Settings,
+	options runtimeclient.ListOptions) (
```
trewest, line 14: One last thing, we actually just made an update so that the options param for list functions is optional. Can you align this function with this example - https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/bmh/list.go#L29-L41

trewest (changes requested):
Just one change and we can merge

trewest (approved):
/lgtm

PR #182, opened 2023-11-08
Added WithVolumeMounts method

yprokule:
Method that allows to configure volume mounts within container.

### DIFF
--- a/pkg/pod/container.go
+++ b/pkg/pod/container.go
@@ -235,6 +235,32 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMount adds a pod volume mount inside the container.
+func (builder *ContainerBuilder) WithVolumeMount(volMount v1.VolumeMount) *ContainerBuilder {
+	glog.V(100).Infof("Adding VolumeMount to the %s container's definition", builder.definition.Name)
+
+	if volMount.Name == "" {
+		glog.V(100).Infof("Container's VolumeMount name cannot be empty")
+
+		builder.errorMsg = "container's volume mount name is empty"
+	}
+
+	if volMount.MountPath == "" {
+		glog.V(100).Infof("Container's VolumeMount mount path cannot be empty")
+
+		builder.errorMsg = "container's volume mount path is empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof("VolumeMount %s will be mounted at %s", volMount.Name, volMount.MountPath)
+	builder.definition.VolumeMounts = append(builder.definition.VolumeMounts, volMount)
+
+	return builder
+}
+
 // GetContainerCfg returns Container struct.
 func (builder *ContainerBuilder) GetContainerCfg() (*v1.Container, error) {
 	glog.V(100).Infof("Returning configuration for container %s", builder.definition.Name)


trewest:
@yprokule No need to add [container] to the commit message. Maybe just update the message to be "Add WithVolumeMounts method to ContainerBuilder"

pkg/pod/container.go
```diff
@@ -235,6 +235,29 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMounts configures pod volumes to mount inside the container.
+func (builder *ContainerBuilder) WithVolumeMounts(volMounts []v1.VolumeMount) *ContainerBuilder {
+	glog.V(100).Infof("Adding VolumeMount(s) to the container's definition")
```
trewest, line 240: Can you add the container name to glog so it is clear?

yprokule:
> @yprokule No need to add [container] to the commit message. Maybe just update the message to be "Add WithVolumeMounts method to ContainerBuilder"

thanks, updated

pkg/pod/container.go
yprokule: thanks, added

yprokule:
Container definition before invoking _WithVolumeMounts_ method:
```bash
&v1.Container{Name:"withmacvlan-c-one", Image:"registry.lab.com:5000/dmeo/snife:latest", Command:[]string{"/bin/sh", "-c", "/opt/net/config.sh 10.1.25.10/26; nc -k -l 10.1.25.10 1111"}, Args:[]string(nil), WorkingDir:"", Ports:[]v1.ContainerPort(nil), EnvFrom:[]v1.EnvFromSource(nil), Env:[]v1.EnvVar(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList(nil), Claims:[]v1.ResourceClaim(nil)},
 VolumeMounts:[]v1.VolumeMount(nil), 
VolumeDevices:[]v1.VolumeDevice(nil), LivenessProbe:(*v1.Probe)(nil), ReadinessProbe:(*v1.Probe)(nil), StartupProbe:(*v1.Probe)(nil), Lifecycle:(*v1.Lifecycle)(nil), TerminationMessagePath:"", TerminationMessagePolicy:"", ImagePullPolicy:"", SecurityContext:(*v1.SecurityContext)(0xc000e38000), Stdin:false, StdinOnce:false, TTY:false}(*v1.Container)
```
and after adding _VolumeMount_:
```go
    fmt.Printf("Setting VolumeMounts")
    volMounts := []v1.VolumeMount{
                    v1.VolumeMount{
                        Name: "configs",
                        MountPath: "/opt/net/",
                        ReadOnly: true,
                    },
                }

    deployContainer = deployContainer.WithVolumeMounts(volMounts)
```
```bash
&v1.Container{Name:"withmacvlan-c-one", Image:"registry.lab.com:5000/dmeo/snife:latest", Command:[]string{"/bin/sh", "-c", "/opt/net/config.sh 10.1.25.10/26; nc -k -l 10.1.25.10 1111"}, Args:[]string(nil), WorkingDir:"", Ports:[]v1.ContainerPort(nil), EnvFrom:[]v1.EnvFromSource(nil), Env:[]v1.EnvVar(nil), Resources:v1.ResourceRequirements{Limits:v1.ResourceList(nil), Requests:v1.ResourceList(nil), Claims:[]v1.ResourceClaim(nil)}, 
   VolumeMounts:[]v1.VolumeMount{v1.VolumeMount{Name:"configs", ReadOnly:true, MountPath:"/opt/net/", SubPath:"", MountPropagation:(*v1.MountPropagationMode)(nil), SubPathExpr:""}},  
VolumeDevices:[]v1.VolumeDevice(nil), LivenessProbe:(*v1.Probe)(nil), ReadinessProbe:(*v1.Probe)(nil), StartupProbe:(*v1.Probe)(nil), Lifecycle:(*v1.Lifecycle)(nil), TerminationMessagePath:"", TerminationMessagePolicy:"", ImagePullPolicy:"", SecurityContext:(*v1.SecurityContext)(0xc000e38000), Stdin:false, StdinOnce:false, TTY:false}(*v1.Container)
```

pkg/pod/container.go
```diff
@@ -235,6 +235,29 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
...
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	if builder.definition.VolumeMounts != nil {
+		builder.definition.VolumeMounts = append(builder.definition.VolumeMounts, volMounts...)
+	} else {
+		builder.definition.VolumeMounts = volMounts
+	}
```
ElaiShalevRH, line 256: This block can maybe be simplified to just
`builder.definition.VolumeMounts = append(builder.definition.VolumeMounts, volMounts...)`
same result even if `builder.definition.VolumeMounts `is nil (unless you want to add logs in the conditional)

pkg/pod/container.go
```diff
@@ -235,6 +235,29 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMounts configures pod volumes to mount inside the container.
+func (builder *ContainerBuilder) WithVolumeMounts(volMounts []v1.VolumeMount) *ContainerBuilder {
```
cdvultur, line 239: @trewest @kononovn Wondering if we want to have builders for volumeMounts. 

Still I see many functions that take in a `v1.VolumeMount` or `[]v1.VolumeMount`

pkg/pod/container.go
```diff
@@ -235,6 +235,29 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
...
+		glog.V(100).Infof("Container's VolumeMounts list is empty")
+
+		builder.errorMsg = "container's volume mount list is empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	if builder.definition.VolumeMounts != nil {
```
cdvultur, line 252: this one could be as @ElaiShalevRH said:

`builder.definition.VolumeMounts = append(builder.definition.VolumeMounts, volMounts...)`

yprokule: thanks, updated

pkg/pod/container.go
trewest: @yprokule Can we change this to `WithVolumeMount`? If you have multiple mounts you can just chain them together (`WithVolumeMount(<def>).WithVolumeMounts(<def>)`)

trewest (changes requested):
Just one comment

pkg/pod/container.go
yprokule: @trewest updated

pkg/pod/container.go
```diff
@@ -235,6 +235,31 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMount configures pod volumes to mount inside the container.
+func (builder *ContainerBuilder) WithVolumeMount(volMount v1.VolumeMount) *ContainerBuilder {
+	glog.V(100).Infof("Adding VolumeMount to the %q container's definition", builder.definition.Name)
```
trewest, line 240: For consistency
```suggestion
	glog.V(100).Infof("Adding VolumeMount to the %s container's definition", builder.definition.Name)
```

yprokule: thanks, updated

trewest (approved):
/lgtm

pkg/pod/container.go
```diff
@@ -235,6 +235,31 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMount configures pod volumes to mount inside the container.
```
achuzhoy, line 238: nit: s'/configures pod volumes/adds a pod volume/'

yprokule: thanks, updated

pkg/pod/container.go
```diff
@@ -235,6 +235,31 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMount  adds a pod volume mount inside the container.
```
cdvultur, line 238: Remove extra space between `WithVolumeMount` and `adds`

pkg/pod/container.go
```diff
@@ -235,6 +235,31 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
...
+func (builder *ContainerBuilder) WithVolumeMount(volMount v1.VolumeMount) *ContainerBuilder {
+	glog.V(100).Infof("Adding VolumeMount to the %s container's definition", builder.definition.Name)
```
cdvultur, line 240: Wondering if it helpful to log the volume name or mount path ? This could be moved after the Name and MountPath checks are done.

Based on the example in the comments, and the fact that the volume it will be created in the test, I would add some more details.

cdvultur (changes requested):
Minor changes

pkg/pod/container.go
yprokule: removed, thanks

pkg/pod/container.go
yprokule: Added logging of `VolumeMount`'s _Name_ and _Path_

cdvultur (approved):
/lgtm

pkg/pod/container.go
```diff
@@ -235,6 +235,32 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMount adds a pod volume mount inside the container.
+func (builder *ContainerBuilder) WithVolumeMount(volMount v1.VolumeMount) *ContainerBuilder {
+	glog.V(100).Infof("Adding VolumeMount to the %s container's definition", builder.definition.Name)
+
+	if volMount.Name == "" {
+		glog.V(100).Infof("Container's VolumeMount name cannot be empty")
```
achuzhoy, line 243: BTW,  we're able technically to mount a volume inside a container without a name. An arbitrary name is generated in such cases. Do we still want to fail here?

yprokule: Thanks @achuzhoy  - unfortunately I've no idea where is that arbitrary name generate(by `oc` client or by some resource schema validation function on the server side, etc) so cannot give U a definitive answer.
I prefer to have it state clearly but if U disagree and  insist on allowing it to be empty - please let me know

trewest: @achuzhoy There must be some linkage between pod.WithVolume, deployment.WithVolume and this method. I think that linkage is the name alone

PR #183, opened 2023-11-09
WithVolume method added to pod pkg

yprokule:
Method that allows to configure _ConfigMap_ volume on the pod

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -599,6 +599,30 @@ func (builder *Builder) WithPrivilegedFlag() *Builder {
 	return builder
 }
 
+// WithVolume attaches given volume to a pod.
+func (builder *Builder) WithVolume(volume v1.Volume) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	if volume.Name == "" {
+		glog.V(100).Infof("The volume's Name cannot be empty")
+
+		builder.errorMsg = "The volume's Name cannot be empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof("Adding volume %s to pod %s in namespace %s",
+		volume.Name, builder.Definition.Name, builder.Definition.Namespace)
+
+	builder.Definition.Spec.Volumes = append(builder.Definition.Spec.Volumes, volume)
+
+	return builder
+}
+
 // WithLocalVolume attaches given volume to all pod's containers.
 func (builder *Builder) WithLocalVolume(volumeName, mountPath string) *Builder {
 	if valid, _ := builder.validate(); !valid {


yprokule:
Tested with next example:
```bash
    var myVolMode int32
    myVolMode = 511

    myPod = myPod.WithConfigMapVolume("configs", "yp-netconfig", &myVolMode)
```

pkg/pod/container.go
```diff
@@ -235,6 +235,25 @@ func (builder *ContainerBuilder) WithEnvVar(name, value string) *ContainerBuilde
 	return builder
 }
 
+// WithVolumeMounts configures pod volumes to mount inside the container.
+func (builder *ContainerBuilder) WithVolumeMounts(volMounts []v1.VolumeMount) *ContainerBuilder {
```
trewest, line 239: Can you change this to ` WithVolumeMount` and just accept a single v1.VolumeMount? Multiple mounts can be chained together.

pkg/pod/pod.go
```diff
@@ -599,6 +599,49 @@ func (builder *Builder) WithPrivilegedFlag() *Builder {
 	return builder
 }
 
+// WithVolume attaches given volume to all pod's containers.
+func (builder *Builder) WithConfigMapVolume(volumeName, cmName string, defMode *int32) *Builder {
```
trewest, line 603: Since Volumes support a large amount of sources, can you just make this generic? Like `WithVolume` that accepts a v1.Volume?

trewest: I think it should be pretty similar to the WithVolumeMount method

trewest (changes requested):
Couple of comments

pkg/pod/container.go
yprokule: resolved in #182

pkg/pod/pod.go
cdvultur: Indeed, from what I know `WithLocalVolume` does use configMap.

What is does not do, is setting the `defaultMode`. 
If that is what you want, changing the signature of this function could break existing usage of that function.

yprokule: No `WithConfigMapVolume` should work only with _ConfigMap_ volume. `WithLocalVolume` does extra things e.g. adding _VolumeMount_ to all the containers or doesn't allow configuration of the _VolumeMount_'s name/path.

Based on @trewest comment I'll check creating of a generic method that will add _Volume_ (despite type) to the _Pod_

yprokule: @trewest updated

yprokule:
tested with next snippet
```go
    myVolume := v1.Volume{
        Name: "configs",
        VolumeSource: v1.VolumeSource{
            ConfigMap: &v1.ConfigMapVolumeSource{
                DefaultMode: &myVolMode,
                LocalObjectReference: v1.LocalObjectReference{
                        Name: "yp-netconfig",
                },
            },
        },
    }

    myPod = myPod.WithVolume(myVolume)
```

pkg/pod/pod.go
```diff
@@ -599,6 +599,29 @@ func (builder *Builder) WithPrivilegedFlag() *Builder {
...
+		glog.V(100).Infof("The volume's Name cannot be empty")
+
+		builder.errorMsg = "The volume's Name cannot be empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof("Configuring volume %q for pod", volume.Name)
```
trewest, line 618:
```suggestion
	glog.V(100).Infof("Adding volume %s to pod %s in namespace %s", volume.Name, builder.Definition.Name, builder.Definition.Namespace)
```

pkg/pod/pod.go
```diff
@@ -599,6 +599,29 @@ func (builder *Builder) WithPrivilegedFlag() *Builder {
 	return builder
 }
 
+// WithVolume attaches given volume to a pod containers.
```
trewest, line 602:
```suggestion
// WithVolume attaches given volume to a pod.
```

trewest (changes requested):
Two small nits, otherwise looks good

pkg/pod/pod.go
yprokule: thanks, updated

pkg/pod/pod.go
yprokule: thanks, updated

trewest (approved):
/lgtm

cdvultur (approved):
/lgtm

PR #184, opened 2023-11-09
WithVolumes method added to Deployment

yprokule:
`WithVolumes` method allows to configure volumes on the _Deployment_ object

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -315,6 +315,32 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
 	return builder
 }
 
+// WithVolume attaches given volume to the deployment.
+func (builder *Builder) WithVolume(deployVolume coreV1.Volume) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	if deployVolume.Name == "" {
+		glog.V(100).Infof("The volume's name cannot be empty")
+
+		builder.errorMsg = "The volume's name cannot be empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof("Adding volume %s to deployment %s in namespace %s",
+		deployVolume.Name, builder.Definition.Name, builder.Definition.Namespace)
+
+	builder.Definition.Spec.Template.Spec.Volumes = append(
+		builder.Definition.Spec.Template.Spec.Volumes,
+		deployVolume)
+
+	return builder
+}
+
 // WithOptions creates deployment with generic mutation options.
 func (builder *Builder) WithOptions(options ...AdditionalOptions) *Builder {
 	if valid, _ := builder.validate(); !valid {


pkg/deployment/deployment.go
```diff
@@ -315,6 +315,32 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
 	return builder
 }
 
+// WithVolumes sets list of volumes that can be mounted by containers belonging to the deployment.
+func (builder *Builder) WithVolumes(deployVolumes []coreV1.Volume) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Adding %d volumes to deployment %s in namespace %s",
```
achuzhoy, line 324: Looks like this glog entry should be moved to after the  len(deployVolumes) == 0 check

trewest:
Hi @yprokule please check this [PR](https://github.com/openshift-kni/eco-goinfra/pull/165/files). It was essentially done besides my last two comments:
https://github.com/openshift-kni/eco-goinfra/pull/165/files#r1385208288
https://github.com/openshift-kni/eco-goinfra/pull/165/files#r1385209332
https://github.com/openshift-kni/eco-goinfra/pull/165/files#r1385208752
Es

pkg/deployment/deployment.go
yprokule: yeah, I was thinking about it as well but it looks harmless  so left it here. please let me know if it should be moved further, thanks

achuzhoy: I would.
glog.V(100).Infof("The volume list of the deployment is empty") would come in case of 0

and "Adding %d volumes to deployment %s in namespace %s" in case > 0

pkg/deployment/deployment.go
```diff
@@ -315,6 +315,32 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
 	return builder
 }
 
+// WithVolumes sets list of volumes that can be mounted by containers belonging to the deployment.
+func (builder *Builder) WithVolumes(deployVolumes []coreV1.Volume) *Builder {
```
trewest, line 319: Can you please change it to WithVolume and accept a single coreV1.Volume? Multiple volumes can be chained together

trewest (changes requested):
One comment

pkg/deployment/deployment.go
yprokule: thanks, updated

pkg/deployment/deployment.go
yprokule: thanks, updated

pkg/deployment/deployment.go
```diff
@@ -315,6 +315,32 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
 	return builder
 }
 
+// WithVolume sets list of volumes that can be mounted by containers belonging to the deployment.
```
trewest, line 318: Please update the comment to reflect that the updated method only adds one volume instead of a list

pkg/deployment/deployment.go
```diff
@@ -315,6 +315,32 @@ func (builder *Builder) WithServiceAccountName(serviceAccountName string) *Build
...
+		glog.V(100).Infof("The volume's name cannot be empty")
+
+		builder.errorMsg = "The volume's name cannot be empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	glog.V(100).Infof("Adding volume %q to deployment %s in namespace %s",
```
trewest, line 334: For consistency
```suggestion
	glog.V(100).Infof("Adding volume %s to deployment %s in namespace %s",
```

trewest (changes requested):
Two nits, otherwise looks good

pkg/deployment/deployment.go
trewest: `WithVolume attaches given volume to a deployment`

yprokule: thanks, updated

pkg/deployment/deployment.go
yprokule: thanks, updated

cdvultur (approved):
/lgtm

trewest (approved):
/lgtm

PR #186, opened 2023-11-15
Added policySet object, updated client

shaior:


trewest (approved):
/lgtm

cdvultur (approved):
/lgtm

Was confused why there is `v1` and `v1beta1` but looking at the CRDs it seems that the policyset is not a `v1`

PR #192, opened 2023-11-19
Added placementrule to ocm pkg

shaior:


policySet.yml
```diff
@@ -0,0 +1,36 @@
+apiVersion: policy.open-cluster-management.io/v1beta1
```
achuzhoy, line 1: Do you plan on actually including this file and policyset.yml ?
If yes - why?

shaior: No, that's probably a leftover i missed, deleted, thanks.

trewest (approved):
/lgtm

PR #202, opened 2023-11-30
pkg: added multinetworkpolicy to networkpolicy pkg

kononovn:


pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+		glog.V(100).Infof("The namespace of the MultiNetworkPolicy is empty")
+
+		builder.errorMsg = "The MultiNetworkPolicy 'namespace' cannot be empty"
+	}
+
+	return builder
+}
+
+// WithPodSelector adds podSelector to MultiNetworkPolicy.
+func (builder *MultiNetworkPolicyBuilder) WithPodSelector(
```
achuzhoy, line 60: This func returns builder anyway.
Do we not want to return errors in case there are errors?

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating MultiNetworkPolicy %s in %s namespace with the annotation defined: %v",
+		builder.Definition.Name, builder.Definition.Namespace, annotation)
+
+	if annotation == "" {
+		glog.V(100).Infof("The annotation can not be empty map")
```
achuzhoy, line 96: empty string?

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	builder.Definition.Annotations = map[string]string{"k8s.v1.cni.cncf.io/policy-for": annotation}
+
+	return builder
+}
+
+// WithEmptyIngress adds empty ingress rule to the MultiNetworkPolicy. Empty ingress denys all.
```
achuzhoy, line 110: denies

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+
+		builder.errorMsg = "The MultiNetworkPolicy 'namespace' cannot be empty"
+	}
+
+	return builder
+}
+
+// WithPodSelector adds podSelector to MultiNetworkPolicy.
+func (builder *MultiNetworkPolicyBuilder) WithPodSelector(
+	podSelector *metav1.LabelSelector) *MultiNetworkPolicyBuilder {
```
trewest, line 61: Is there a reason you are using *metav1.LabelSelector here? If you just use metav1.LabelSelector, you will not need to check if it is nil nor dereference it on line 80

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+		return builder
+	}
+
+	builder.Definition.Spec.PodSelector = *podSelector
+
+	return builder
+}
+
+// WithAnnotation adds Annotation to the MultiNetworkPolicy.
+func (builder *MultiNetworkPolicyBuilder) WithAnnotation(annotation string) *MultiNetworkPolicyBuilder {
```
trewest, line 86: Personally, I think this method name is a bit misleading. It doesn't accept just any user-defined annotation but only an annotation value with k8s.v1.cni.cncf.io/policy-for as the key.  I think it would be more appropriate to name it something like `WithPolicyForAnnotation` or something similar that you think sufficiently conveys what this method does.

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+		glog.V(100).Infof("The annotation can not be empty map")
+
+		builder.errorMsg = "The annotation is an empty string"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	builder.Definition.Annotations = map[string]string{"k8s.v1.cni.cncf.io/policy-for": annotation}
```
trewest, line 105: Rather than overwriting all of the annotations, I think this is better
```suggestion
        if builder.Definition.Annotations == nil {
           builder.Definition.Annotations = make(map[string]string)
        }
        
        
	builder.Definition.Annotations["k8s.v1.cni.cncf.io/policy-for"] = annotation
```

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+func (builder *MultiNetworkPolicyBuilder) WithEmptyIngress() *MultiNetworkPolicyBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating MultiNetworkPolicy %s in %s namespace with the empty Ingress rule deny all",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if builder.errorMsg != "" {
```
trewest, line 120: This is redundant with the builder.validate() block

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+	ingressRule v1beta1.MultiNetworkPolicyIngressRule) *MultiNetworkPolicyBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Creating multiNetworkPolicy %s in %s namespace with the Ingress rule defined: %v",
+		builder.Definition.Name, builder.Definition.Namespace, ingressRule)
+
+	if builder.errorMsg != "" {
```
trewest, line 140: This is redundant with the builder.validate() block

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+
+	glog.V(100).Infof(
+		"Creating multiNetworkPolicy %s in %s namespace with the Policy type defined: %v",
+		builder.Definition.Name, builder.Definition.Namespace, policyType)
+
+	if policyType == "" {
+		glog.V(100).Infof("The policy type can not be empty map")
+
+		builder.errorMsg = "The policy Type is an empty map"
```
trewest, line 162:
```suggestion
		builder.errorMsg = "The policy Type is an empty string"
```

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+func (builder *MultiNetworkPolicyBuilder) Delete() error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof("Deleting the MultiNetworkPolicy object %s from %s namespace",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if !builder.Exists() {
+		return fmt.Errorf("MultiNetworkPolicy cannot be deleted because it does not exist")
```
trewest, line 262: Go standard is for errors to begin with a lowercase letter

trewest (changes requested):
Few comments

pkg/networkpolicy/multinetworkpolicy.go
kononovn: We don't return error from builder functions

pkg/networkpolicy/multinetworkpolicy.go
kononovn: Refactored

pkg/networkpolicy/multinetworkpolicy.go
```diff
@@ -0,0 +1,328 @@
...
+		builder.errorMsg = "MultiNetworkPolicy 'name' cannot be empty"
+	}
+
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the MultiNetworkPolicy is empty")
+
+		builder.errorMsg = "MultiNetworkPolicy 'namespace' cannot be empty"
+	}
+
+	if builder.errorMsg != "" {
```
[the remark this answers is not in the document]
kononovn: We can think about improvement here.

trewest (approved):
/lgtm

trewest (approved):
/lgtm

PR #207, opened 2023-12-06
Method WithToleration added to deployment pkg

yprokule:
Method that allows to set toleration on the _Deployment_


trewest (approved):
/lgtm

PR #217, opened 2024-01-05
Kmm ManagedClusterModule support

cdvultur:


pkg/kmm/managedclustermodule.go
```diff
@@ -0,0 +1,301 @@
...
+
+// WithSelector set the selector for the managedclustermodule object.
+func (builder *ManagedClusterModuleBuilder) WithSelector(
+	selector map[string]string) *ManagedClusterModuleBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	if selector == nil {
+		builder.errorMsg = "invalid 'selector' argument can not be nil"
```
trewest, line 100:
```suggestion
	if len(selector) == 0 {
		builder.errorMsg = "invalid 'selector' argument cannot be empty"
```

pkg/kmm/managedclustermodule.go
```diff
@@ -0,0 +1,301 @@
...
+		if err == nil {
+			builder.Object = builder.Definition
+		}
+	}
+
+	return builder, err
+}
+
+// Update modifies an existing manageclustermodule on the cluster.
+func (builder *ManagedClusterModuleBuilder) Update() (*ManagedClusterModuleBuilder, error) {
```
trewest, line 195: Does this work correctly without the force argument?

trewest (changes requested):
Few comments

pkg/kmm/managedclustermodule.go
```diff
@@ -0,0 +1,301 @@
...
+
+// WithSpokeNamespace sets the namespace where the module will be deployed on the spoke.
+func (builder *ManagedClusterModuleBuilder) WithSpokeNamespace(
+	spokeNamespace string) *ManagedClusterModuleBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	if spokeNamespace == "" {
+		builder.errorMsg = "invalid 'module' argument can not be nil"
```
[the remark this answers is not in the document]
cdvultur: Updated. Thanks !

pkg/kmm/managedclustermodule.go
cdvultur: I am reusing the same from the KMM Module and it is working as it should. 
Still this specific method did not tested it yet.

pkg/kmm/managedclustermodule.go
```diff
@@ -0,0 +1,301 @@
...
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	builder.Definition.Spec.SpokeNamespace = spokeNamespace
+
+	return builder
+}
+
+// WithSelector set the selector for the managedclustermodule object.
```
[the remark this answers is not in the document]
cdvultur: Fixed! Thanks !

pkg/kmm/managedclustermodule.go
trewest: I think you should at least add this block to the end of the method:
```
if err == nil {
}
```
Otherwise, I don't see how builder.Object is being updated after a successful request. Here is an example from another resource that doesn't have a versioned client interface - https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/hive/clusterdeployment.go#L290-292

cdvultur: You are right. 

In my case I was updating the object, but I was checking that the operator did the updates. 
And those checks are done with `dmesg`/`lsmod`, for which I did not had to access the `builder.Object `

Updated the code. Thanks !

achuzhoy (approved):

trewest (approved):
/lgtm

PR #226, opened 2024-01-18
Method to set PVC VolumeMode

yprokule:
Added method that defines what type of volume is required by the claim.

### DIFF
--- a/pkg/storage/pvc.go
+++ b/pkg/storage/pvc.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/golang/glog"
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
 	"github.com/openshift-kni/eco-goinfra/pkg/msg"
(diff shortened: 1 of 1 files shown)


pkg/storage/pvc.go
```diff
@@ -315,3 +344,17 @@ func (builder *PVCBuilder) validate() (bool, error) {
 
 	return true, nil
 }
+
+// validateVolumeMode validates if requested volume mode is valid for PVC.
+func validateVolumeMode(volumeMode string) bool {
+	glog.V(100).Info("Validating volumeMode %s", volumeMode)
+
+	var validVolumeModes = map[string]string{
```
achuzhoy, line 352: do we need to create a dict for 2 possible string values? i.e. why not just comparing to one of the strings?

pkg/storage/pvc.go
```diff
@@ -150,6 +150,35 @@ func (builder *PVCBuilder) WithStorageClass(storageClass string) (*PVCBuilder, e
...
+		builder.errorMsg = fmt.Sprintf("Empty volumeMode requested for the PersistentVolumeClaim %s",
+			builder.Definition.Name)
+
+		return builder, fmt.Errorf(builder.errorMsg)
+	}
+
+	if !validateVolumeMode(volumeMode) {
+		glog.V(100).Infof(fmt.Sprintf("Unsupported VolumeMode: %s", volumeMode))
+
+		builder.errorMsg = fmt.Sprintf("Unsupoorted VolumeMode %q requested for %s PersistentVolumeClaim",
```
achuzhoy, line 169: misspell: Unsupoorted
Add the namespace

pkg/storage/pvc.go
```diff
@@ -150,6 +150,35 @@ func (builder *PVCBuilder) WithStorageClass(storageClass string) (*PVCBuilder, e
...
+			volumeMode, builder.Definition.Name)
+
+		return builder, fmt.Errorf(builder.errorMsg)
+	}
+
+	nVolMode := v1.PersistentVolumeMode(volumeMode)
```
achuzhoy, line 175: what is n is nVolMode ?

pkg/storage/pvc.go
trewest: Or just use the slices.Contains() function. Example - https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/rbac/clusterrolebinding.go#L85

pkg/storage/pvc.go
yprokule: normalized 😸 `builder.Definition.Spec.VolumeMode` is of type `*v1.PersistentVolumeMode` so U cannot just do conversion on the fly but have to use intermediary variable

pkg/storage/pvc.go
yprokule: thanks, updated

pkg/storage/pvc.go
```diff
@@ -150,6 +150,35 @@ func (builder *PVCBuilder) WithStorageClass(storageClass string) (*PVCBuilder, e
...
 }
 
+// WithVolumeMode configures what type of volume is required by the claim.
+func (builder *PVCBuilder) WithVolumeMode(volumeMode string) (*PVCBuilder, error) {
+	glog.V(100).Infof("Set VolumeMode %s for the PersistentVolumeClaim", volumeMode)
+
+	if volumeMode == "" {
+		glog.V(100).Infof("Empty volumeMode requested for the PersistentVolumeClaim")
+
+		builder.errorMsg = fmt.Sprintf("Empty volumeMode requested for the PersistentVolumeClaim %s",
```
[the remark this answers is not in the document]
yprokule: sure, added

pkg/storage/pvc.go
yprokule: thanks, updated

pkg/storage/pvc.go
achuzhoy: imho would be more beneficial to have a meaningful name,but up to you.

trewest (approved):
/lgtm - pending updates for Sasha's comment

pkg/storage/pvc.go
yprokule: updated and added comment

achuzhoy (approved):

cdvultur (approved):
/lgtm

PR #229, opened 2024-01-25
Updated client.go. Added K8sClient to clients struct

yprokule:
Methods for nodes cordoning, uncordoning and draining added


pkg/nodes/drain.go
```diff
@@ -0,0 +1,67 @@
...
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/kubectl/pkg/drain"
+)
+
+// NewDrainHelperBuilder builds drain Helper that contains parameters to control the behaviour of drain.
+func NewDrainHelperBuilder(
```
kononovn, line 16: I think:
Drain/Cordon/Uncordon functions have to a part of node.Builder struct.
Please add these methods as part of Builder in node.go file

kononovn (changes requested):
Please check comment above

kononovn (changes requested):
PR looks good. It's almost ready to merge.
Can you please do following changes:

Lave 2 commits instead of 3 commits under this PR.
 
The first commit has to have following message:
"Bump dependence. Add kubectl drain lib"
Second commit message:
"Updated client.go. Added K8sClient to clients struct."

trewest (approved):
/lgtm

PR #232, opened 2024-01-29
Drain/Cordon/Uncordon methods for Node object

yprokule:


yprokule:
this one depends on #229 so has changes from it here. when #229 merged this one is rebased and will contain only changes to the `pkg/nodes/node.go` only

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
-	apiClient  *clients.Settings
-	errorMsg   string
+	Definition  *v1.Node
+	Object      *v1.Node
+	apiClient   *clients.Settings
+	errorMsg    string
+	drainHelper *drain.Helper
+}
+
+// NewDrainHelperbuilds drain Helper that contains parameters to control the behaviour of drain.
```
achuzhoy, line 35: NewDrainHelper

yprokule: thanks, updated

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
+	ignoreDaemonsets bool,
+	gracePeriod int,
+	timeout time.Duration,
+	skipWaitForDeleteTimeoutSeconds int,
+	deleteLocalData bool) *Builder {
+	glog.V(100).Infof("Creating new DrainOptions config")
+	glog.V(100).Infof(
+		fmt.Sprintf("Creating new DrainOptions force: %v", force))
+	glog.V(100).Infof(
+		fmt.Sprintf("Creating new DrainOptions gracePeriod: %d seconds", gracePeriod))
```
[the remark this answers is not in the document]
yprokule: thanks, updated

kononovn (changes requested):
@yprokule Please re-base

yprokule:
> @yprokule Please re-base

@kononovn rebased

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
-	errorMsg   string
+	Definition  *v1.Node
+	Object      *v1.Node
+	apiClient   *clients.Settings
+	errorMsg    string
+	drainHelper *drain.Helper
+}
+
+// NewDrainHelper builds drain Helper that contains parameters to control the behaviour of drain.
+func (builder *Builder) NewDrainHelper(
```
kononovn, line 36: Let's change a bit API.
I would recommend to Define following function:

SetDrainHelper which'll actually configure drainHelper *drain.Helper object based on given parameters.
setDefaultDrainHelper - private function that configures *drain.Helper based on some "best practices" configuration.

No need to add "node" for the following public functions because the object itself is a node.
Drain
Cordon
Uncordon

For the functions above please add simple condition.
if builder.drainHelper == nil{
    setDefaultDrainHelper()
}

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
+	force bool,
```
kononovn, line 37: Please keep arguments of the same type close to each other.

kononovn (changes requested):
@yprokule You PR looks good. Can you please check few comments above? I think we can make it even better.

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
+}
+
+// DrainNode drains the node.
+func (builder *Builder) DrainNode() error {
+	glog.V(100).Infof(fmt.Sprintf("Draining node %s", builder.Definition.Name))
+
+	return drain.RunNodeDrain(builder.drainHelper, builder.Definition.Name)
+}
+
+// CordonNode cordon node.
```
achuzhoy, line 87: cordons node

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
+func (builder *Builder) CordonNode() error {
+	glog.V(100).Infof(fmt.Sprintf("Cordoning node %s", builder.Definition.Name))
+
+	return drain.RunCordonOrUncordon(builder.drainHelper, builder.Definition, true)
+}
+
+// UncordoneNode uncordons node.
+func (builder *Builder) UncordonNode() error {
+	glog.V(100).Infof(fmt.Sprintf("Uncordonning node %s", builder.Definition.Name))
```
achuzhoy, line 96: Uncordoning (single n)

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
+		Force: force,
+		// GracePeriodSeconds is how long to wait for a pod to terminate.
+		GracePeriodSeconds: gracePeriod,
+		// Ignore DaemonSet-managed pods
+		IgnoreAllDaemonSets: ignoreDaemonsets,
+		// The length of time to wait before giving up
+		Timeout: timeout,
+		// Local data from emptyDir volumes will be deleted
+		DeleteEmptyDirData: deleteLocalData,
+		// when the node is drained
```
achuzhoy, line 70: need a couple more words. Not clear

pkg/nodes/node.go
yprokule: `NewDrainHelper` renamed to `SetDrainHelper`

WRT `setDefaultDrainHelper` I'm not sure where to plug it in as there's no a method that create _Node_ object, only the one that pulls it in or the one that lists nodes. Could U please elaborate on how to achieve it? Thanks!

pkg/nodes/node.go
yprokule: updated, thanks

pkg/nodes/node.go
```diff
@@ -23,10 +25,77 @@ const (
...
+
+// NewDrainHelper builds drain Helper that contains parameters to control the behaviour of drain.
+func (builder *Builder) NewDrainHelper(
+	force bool,
+	ignoreDaemonsets bool,
+	gracePeriod int,
+	timeout time.Duration,
+	skipWaitForDeleteTimeoutSeconds int,
+	deleteLocalData bool) *Builder {
+	glog.V(100).Infof("Creating new DrainOptions config")
```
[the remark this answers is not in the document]
yprokule: updated to one line

pkg/nodes/node.go
yprokule: updated

pkg/nodes/node.go
yprokule: bad split, that's related to the previous line, updated

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+		Out:                os.Stdout,
+		ErrOut:             os.Stderr,
+		// If pod DeletionTimestamp older than N seconds, skip waiting for the pod.
+		SkipWaitForDeleteTimeoutSeconds: skipWaitForDeleteTimeoutSeconds,
+	}
+
+	return builder
+}
+
+// Drain drains the node.
```
achuzhoy, line 78:
```suggestion
// Drain evacuates pods from the node.
```

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+func (builder *Builder) Drain() error {
+	glog.V(100).Infof(fmt.Sprintf("Draining node %s", builder.Definition.Name))
+
+	return drain.RunNodeDrain(builder.drainHelper, builder.Definition.Name)
+}
+
+// Cordon cordons node.
```
achuzhoy, line 85:
```suggestion
// Cordon marks the node undeschedulable.
```

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+func (builder *Builder) Cordon() error {
+	glog.V(100).Infof(fmt.Sprintf("Cordoning node %s", builder.Definition.Name))
+
+	return drain.RunCordonOrUncordon(builder.drainHelper, builder.Definition, true)
+}
+
+// Uncordon uncordons node.
```
achuzhoy, line 92:
```suggestion
// Uncordon marks the node schedulable.
```

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+		ErrOut:             os.Stderr,
+		// If pod DeletionTimestamp older than N seconds, skip waiting for the pod.
+		SkipWaitForDeleteTimeoutSeconds: skipWaitForDeleteTimeoutSeconds,
+	}
+
+	return builder
+}
+
+// Drain drains the node.
+func (builder *Builder) Drain() error {
```
kononovn, line 79: Please check my code example below. I think it could help.
```
func (builder *Builder) ensureDrainHelperIsSet() {
	if builder.drainHelper == nil {
		glog.V(100).Infof(fmt.Sprintf("DrainHelper is not initialized for node %s. Init DrainHelper with defaul parameters", builder.Definition.Name))
		builder.SetDrainHelper(false, false, false, 10, 10, 10*time.Minute)
	}
}

func (builder *Builder) Cordon() error {
		return err
	}
	
	builder.ensureDrainHelperIsSet()
	
	glog.V(100).Infof(fmt.Sprintf("Cordoning node %s", builder.Definition.Name))

	return drain.RunCordonOrUncordon(builder.drainHelper, builder.Definition, true)
}
```

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+	glog.V(100).Infof(fmt.Sprintf("Draining node %s", builder.Definition.Name))
```
kononovn, line 80: Please remove fmt.Sprintf. glog function > Infof already supports "formatters"

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+
+	return drain.RunNodeDrain(builder.drainHelper, builder.Definition.Name)
+}
+
+// Cordon cordons node.
+func (builder *Builder) Cordon() error {
+	glog.V(100).Infof(fmt.Sprintf("Cordoning node %s", builder.Definition.Name))
```
kononovn, line 87: Please remove fmt.Sprintf wrapper. Infof already support "formatters"

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+
+	return drain.RunCordonOrUncordon(builder.drainHelper, builder.Definition, true)
+}
+
+// Uncordon marks the node schedulable.
+func (builder *Builder) Uncordon() error {
+	glog.V(100).Infof(fmt.Sprintf("Uncordoning node %s", builder.Definition.Name))
```
kononovn, line 94: Please remove fmt.Sprintf.

pkg/nodes/node.go
```diff
@@ -23,10 +25,75 @@ const (
...
+
+// SetDrainHelper builds drain Helper that contains parameters to control the behaviour of drain.
+func (builder *Builder) SetDrainHelper(
+	force bool,
+	ignoreDaemonsets bool,
+	deleteLocalData bool,
+	gracePeriod int,
+	skipWaitForDeleteTimeoutSeconds int,
+	timeout time.Duration,
+) *Builder {
```
kononovn, line 43: Please remove *Builder from function return. No need for that. This method just sets **builder.drainHelper**

pkg/nodes/node.go
kononovn: @yprokule Please implement Uncordon and Drain using the same approach.

kononovn (changes requested):
Please check my comments above.

pkg/nodes/node.go
yprokule: That's not 100% accurate as evacuation of pods is one of the strategies and the other one is simply to delete pods

achuzhoy: evacuates or deletes (as long as it's not the same word).
I just looked for the first verb here: https://docs.openshift.com/container-platform/4.14/nodes/nodes/nodes-nodes-working.html

pkg/nodes/node.go
yprokule: thanks, updated

pkg/nodes/node.go
yprokule: updated, thanks

pkg/nodes/node.go
yprokule: updated, thanks

pkg/nodes/node.go
yprokule: updated, thanks

pkg/nodes/node.go
yprokule: updated, thanks

pkg/nodes/node.go
yprokule: does updated message work for U?

pkg/nodes/node.go
```diff
@@ -23,10 +25,98 @@ const (
...
+		DeleteEmptyDirData: deleteLocalData,
+		Out:                os.Stdout,
+		ErrOut:             os.Stderr,
+		// If pod DeletionTimestamp older than N seconds, skip waiting for the pod.
+		SkipWaitForDeleteTimeoutSeconds: skipWaitForDeleteTimeoutSeconds,
+	}
+}
+
+// ensureDrainHelperIsSet ensures that drainHelper is always set.
+func (builder *Builder) ensureDrainHelperIsSet() {
```
kononovn, line 77: Private functions have to be placed in the end of the file after public functions.

pkg/nodes/node.go
```diff
@@ -23,10 +25,98 @@ const (
...
+	if builder.drainHelper == nil {
+		glog.V(100).Infof(
+			fmt.Sprintf("DrainHelper is not initialized for node %s. Init DrainHelper with defaul parameters",
+				builder.Definition.Name))
+		builder.SetDrainHelper(false, false, false, 10, 10, 10*time.Minute)
```
kononovn, line 82: Can you please analyse what are the most generic parameters that could work for most of the use-cases?
These params are going to be used by default so it would be good to have most generic settings here.

pkg/nodes/node.go
```diff
@@ -23,10 +25,98 @@ const (
...
+		// If pod DeletionTimestamp older than N seconds, skip waiting for the pod.
+		SkipWaitForDeleteTimeoutSeconds: skipWaitForDeleteTimeoutSeconds,
+	}
+}
+
+// ensureDrainHelperIsSet ensures that drainHelper is always set.
+func (builder *Builder) ensureDrainHelperIsSet() {
+	if builder.drainHelper == nil {
+		glog.V(100).Infof(
+			fmt.Sprintf("DrainHelper is not initialized for node %s. Init DrainHelper with defaul parameters",
```
kononovn, line 80: Please remove fmt.Sprintf

kononovn (changes requested):
Please check few more comments.

pkg/nodes/node.go
yprokule: OK. It's not one solution fits all scenario but updated params should work. Also could be adjusted if there's more data available.

pkg/nodes/node.go
yprokule: done

pkg/nodes/node.go
yprokule: done

kononovn (approved):
/lgtm

achuzhoy (approved):

PR #241, opened 2024-02-05
Update Go to v1.21.6

sebrandon1:
Similar to: 
- https://gitlab.cee.redhat.com/cnf/cnf-gotests/-/merge_requests/657
- https://github.com/openshift-kni/eco-gosystem/pull/64


trewest:
Can you please provide some input as to why you need to update the go version of the project? This infra project is used by a large number of groups working on various openshift components so we are typically only updating if there is justification to. Additionally, our main branches are currently inline with OCP release-4.15 which is currently using go 1.20

PR #245, opened 2024-02-07
lca - ibu: additionionalImages option is added.

achuzhoy:
This will allow providing a configmap reference with additional images.

### DIFF
--- a/pkg/lca/imagebasedupgrade.go
+++ b/pkg/lca/imagebasedupgrade.go
@@ -223,6 +223,22 @@ func (builder *ImageBasedUpgradeBuilder) WithSeedImage(
 	return builder
 }
 
+// WithAdditionalImages adds additionalImages to be used by the imagebasedupgrade.
+func (builder *ImageBasedUpgradeBuilder) WithAdditionalImages(
+	additionalImagesConfigMapName, additionalImagesConfigMapNamespace string) *ImageBasedUpgradeBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Setting additionalImages configmap name %s in namespace %s in the imagebasedupgrade",
+		additionalImagesConfigMapName, additionalImagesConfigMapNamespace)
+
+	builder.Definition.Spec.AdditionalImages =
+		lcav1alpha1.ConfigMapRef{Name: additionalImagesConfigMapName, Namespace: additionalImagesConfigMapNamespace}
+
+	return builder
+}
+
 // WithExtraManifests adds extraManifests to be used by the imagebasedupgrade.
 // This is used to create/configure resources during upgrade.
 func (builder *ImageBasedUpgradeBuilder) WithExtraManifests(


pkg/lca/imagebasedupgrade.go
```diff
@@ -223,6 +223,22 @@ func (builder *ImageBasedUpgradeBuilder) WithSeedImage(
...
 }
 
+// WithAdditionalImages adds additionalImages to be used by the imagebasedupgrade.
+func (builder *ImageBasedUpgradeBuilder) WithAdditionalImages(
+	additionalImagesConfigMapName, additionalImagesConfigMapNamespace string) *ImageBasedUpgradeBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Appending additionalImages's configmap name %s in namespace %s to the imagebasedupgrade",
```
trewest, line 233:
```suggestion
	glog.V(100).Infof("Appending additionalImages configmap name %s in namespace %s to the imagebasedupgrade",
```

trewest (changes requested):
Looks good just one  nit

pkg/lca/imagebasedupgrade.go
```diff
@@ -223,6 +223,22 @@ func (builder *ImageBasedUpgradeBuilder) WithSeedImage(
 	return builder
 }
 
+// WithAdditionalImages adds additionalImages to be used by the imagebasedupgrade.
+func (builder *ImageBasedUpgradeBuilder) WithAdditionalImages(
+	additionalImagesConfigMapName, additionalImagesConfigMapNamespace string) *ImageBasedUpgradeBuilder {
```
kononovn, line 228: Should we validate that string parameters are not empty?

achuzhoy: No need, patching with empty results in:

spec:
  additionalImages:
    name: ""
    namespace: ""

pkg/lca/imagebasedupgrade.go
achuzhoy: Updated. Thanks.

cdvultur (dismissed):
/lgtm

pkg/lca/imagebasedupgrade.go
```diff
@@ -223,6 +223,22 @@ func (builder *ImageBasedUpgradeBuilder) WithSeedImage(
...
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Appending additionalImages configmap name %s in namespace %s to the imagebasedupgrade",
```
trewest, line 233: Just one technicality 
```suggestion
	glog.V(100).Infof("Adding additionalImages configmap name %s in namespace %s to the imagebasedupgrade",
```
Appending make it seem that it is part of a slice. But it is really just one configmap

achuzhoy: You're right.
Thanks.

kononovn (approved):
/lgtm

PR #249, opened 2024-02-15
pod pkg: added WithToleration,NodeSelector,CustomResourcesLimits funcs

kononovn:


pkg/pod/container.go
```diff
@@ -280,6 +280,25 @@ func (builder *ContainerBuilder) WithResourceRequest(hugePages, memory string, c
 	return builder
 }
 
+// WithCustomResourcesLimits applies custom resource limit struct on container.
+func (builder *ContainerBuilder) WithCustomResourcesLimits(resourceList v1.ResourceList) *ContainerBuilder {
+	glog.V(100).Infof("Applying custom resource limit to container: %v", resourceList)
+
+	if resourceList == nil {
```
trewest, line 287: I think it's safer to do len(resourceList) == 0

pkg/pod/container.go
```diff
@@ -280,6 +280,25 @@ func (builder *ContainerBuilder) WithResourceRequest(hugePages, memory string, c
...
+		glog.V(100).Infof("Container's resource limit var 'resourceList' is empty")
+
+		builder.errorMsg = "container's resource limit var 'resourceList' is empty"
+	}
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	builder.definition.Resources.Limits = resourceList
```
cdvultur, line 297: not related to this PR, but the ContainerBuilder struct has `definition` instead of `Definition`

pkg/pod/pod.go
```diff
@@ -577,6 +577,52 @@ func (builder *Builder) WithTolerationToMaster() *Builder {
...
+	return builder
+}
+
+// WithNodeSelector adds a nodeSelector configuration inside the pod.
+func (builder *Builder) WithNodeSelector(nodeSelector map[string]string) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Redefining pod's %s with nodeSelector %v", builder.Definition.Name, nodeSelector)
```
trewest, line 605:
```suggestion
	glog.V(100).Infof("Redefining pod %s in namespace %s with nodeSelector %v", builder.Definition.Name, builder.Definition.Namespace, nodeSelector)
```

pkg/pod/pod.go
```diff
@@ -577,6 +577,52 @@ func (builder *Builder) WithTolerationToMaster() *Builder {
...
+
+	glog.V(100).Infof("Redefining pod's %s with toleration %v", builder.Definition.Name, toleration)
+
+	builder.isMutationAllowed("custom toleration")
+
+	if builder.errorMsg != "" {
+		return builder
+	}
+
+	builder.Definition.Spec.Tolerations = append(builder.Definition.Spec.Tolerations, toleration)
```
cdvultur, line 594: Given that we `append()`, maybe the glog above should say that. Redefine seems to mean _replace_ in most functions. 

Also if we want to append, maybe we should log the existing tolerations?

cdvultur (commented):
One comment

pkg/pod/container.go
kononovn: Yes, I did it intentionally. container.builder is private struct that it not exposed

pkg/pod/pod.go
kononovn: Updated comment

pkg/pod/container.go
cdvultur: Thanks for clarification

cdvultur (approved):
/lgtm

pkg/pod/pod.go
```diff
@@ -577,6 +577,53 @@ func (builder *Builder) WithTolerationToMaster() *Builder {
 	return builder
 }
 
+// WithToleration adds a toleration configuration inside the pod.
+func (builder *Builder) WithToleration(toleration v1.Toleration) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Updating pod's %s with toleration %v", builder.Definition.Name, toleration)
```
achuzhoy, line 586: pod (not pod's)

trewest (dismissed):
/lgtm

PR #251, opened 2024-02-16
IBU: Toggle the automated rollback for postreboot

achuzhoy:
This commit adds the ability to disable the automated rollback in case of failure for post reboot phase. The module was updated too.


achuzhoy:
cc @trewest @kononovn

pkg/lca/imagebasedupgrade.go
```diff
@@ -273,6 +273,21 @@ func (builder *ImageBasedUpgradeBuilder) WithOadpContent(
 	return builder
 }
 
+// AutoRollbackOnFailureDisableForPostReboot allows controlling
+// AutoRollback on failure for post reboot stage. Enabled by default.
+func (builder *ImageBasedUpgradeBuilder) AutoRollbackOnFailureDisableForPostReboot(
```
kononovn, line 278: Please change function name to WithAutoRollbackOnFailureFlag.

pkg/lca/imagebasedupgrade.go
```diff
@@ -273,6 +273,21 @@ func (builder *ImageBasedUpgradeBuilder) WithOadpContent(
...
+	flag bool) *ImageBasedUpgradeBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Setting Auto Rollback on failure for post reboot to %b in imagebasedupgrade", !flag)
```
kononovn, line 284: Please update log to>
Setting Auto Rollback on failure flag for post reboot to %b in imagebasedupgrade

pkg/lca/imagebasedupgrade.go
trewest: How about `WithRollbackDisabledForPostRebootConfig` ? I think we can just remove auto because there isn't anything in the spec not related to auto.

achuzhoy: disclaimer - I spent some time changing the function names yesterday.
The reason I went with AutoRollbackOnFailure in front because of this:
`RESOURCE: autoRollbackOnFailure <Object>

DESCRIPTION:
     <empty>

FIELDS:
   disabledForPostRebootConfig  <boolean>

   disabledForUpgradeCompletion <boolean>

   disabledInitMonitor  <boolean>

   initMonitorTimeoutSeconds    <integer>
`
Thus there will be 3 more functions for AutoRollbackOnFailure.
WDYT?

pkg/lca/imagebasedupgrade.go
achuzhoy: there's an inverse logic - note that I'm inversing the flag's value.
"disableautorolback" it's disabled by default, which means it's enabled

kononovn (approved):
/lgtm

cdvultur (approved):
/lgtm

PR #253, opened 2024-02-22
secret pkg: added Update func plus minor comment change

cdvultur:
Secret pkg seems that it misses the Update func. 
- Adding it with this PR . 
- Update WithData glog comment


achuzhoy (approved):

kononovn (approved):
/lgtm

PR #259, opened 2024-03-01
IBU: Toggle the InitMonitor for automated rollback

achuzhoy:
This commit adds the ability to disable the LCA Init Monitor, which is a post-reboot watchdog that triggers a rollback if the upgrade is not completed within the configured timeout.

### DIFF
--- a/pkg/lca/imagebasedupgrade.go
+++ b/pkg/lca/imagebasedupgrade.go
@@ -289,6 +289,22 @@ func (builder *ImageBasedUpgradeBuilder) AutoRollbackOnFailureInitMonitorTimeout
 	return builder
 }
 
+// AutoRollbackOnFailureDisabledInitMonitor allows disabling the watchdog
+// triggering a rollback upon upgrade failure within the set timeout.
+// Set to false by default.
+func (builder *ImageBasedUpgradeBuilder) AutoRollbackOnFailureDisabledInitMonitor(
+	flag bool) *ImageBasedUpgradeBuilder {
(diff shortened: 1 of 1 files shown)


pkg/lca/imagebasedupgrade.go
```diff
@@ -289,6 +289,22 @@ func (builder *ImageBasedUpgradeBuilder) AutoRollbackOnFailureInitMonitorTimeout
...
+// AutoRollbackOnFailureDisabledInitMonitor allows disabling the watchdog
+// triggering a rollback upon upgrade failure within the set timeout.
+// Set to false by default.
+func (builder *ImageBasedUpgradeBuilder) AutoRollbackOnFailureDisabledInitMonitor(
+	flag bool) *ImageBasedUpgradeBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof("Setting the Init Monitor for Auto Rollback on failure to %b in imagebasedupgrade", !flag)
```
trewest, line 301: Now you're really trying to make me scratch my head 😆

achuzhoy: Exactly the pain we all share with this :)

So the log would be:
"Setting the Init Monitor for Auto Rollback on failure to true in imagebasedupgrade" if we set
Spec.AutoRollbackOnFailure.DisabledInitMonitor to false

trewest (approved):
/lgtm

kononovn (approved):
/lgtm

PR #263, opened 2024-03-08
sriov pkg: Added SriovOperatorConfig

evgenLevin:


pkg/sriov/operatorconfig.go
```diff
@@ -0,0 +1,146 @@
...
+	Definition *srIovV1.SriovOperatorConfig
+	// Created SriovOperatorConfig object.
+	Object *srIovV1.SriovOperatorConfig
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
+	errorMsg  string
+}
+
+// PullOperatorConfig loads an existing SriovOperatorConfig into OperatorConfigBuilder struct.
+func PullOperatorConfig(apiClient *clients.Settings, nsname string) (*OperatorConfigBuilder, error) {
```
cdvultur, line 33: Wonder if there can be different operator configs. In that case this should be `PullDefaultOperatorConfig`

And in case there can be different operator configs, that can be accessed with `PullOperatorConfigByName`.

cdvultur (commented):
Small comment on the config. 

Otherwise lgtm

pkg/sriov/operatorconfig.go
evgenLevin: I don't think we can, it's a defined CRD: srIovV1.SriovOperatorConfig

pkg/sriov/operatorconfig.go
```diff
@@ -0,0 +1,146 @@
...
+				Name:      sriovOperatorConfigName,
+				Namespace: nsname,
+			},
+		},
+	}
+
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the SriovOperatorConfig is empty")
+
+		builder.errorMsg = "metallb 'nsname' cannot be empty"
```
kononovn, line 49: Please fix error message.

kononovn (changes requested):
Please check few comments above

pkg/sriov/operatorconfig.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	if builder.apiClient == nil {
+		glog.V(100).Infof("The %s builder apiclient is nil", resourceCRD)
+
+		return false, fmt.Errorf(fmt.Sprintf("%s builder cannot have nil apiClient", resourceCRD))
+	}
+
+	return true, nil
+}
```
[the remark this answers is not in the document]
evgenLevin: added

pkg/sriov/operatorconfig.go
evgenLevin: fixed

kononovn (approved):
/lgtm

PR #269, opened 2024-03-12
backport: added webhook pkg, updated sriov and nmstate pkgs

evgenLevin:
backport: added webhook pkg, updated sriov and nmstate pkgs


kononovn (approved):
/lgtm

PR #261, opened 2024-03-06
Add path for running unit-tests in Makefile

sebrandon1:
Adds the path `make test` to the Makefile and to the Github workflow.

Lays the groundwork for being able to run unit tests locally and in the CI.

PR #239, opened 2024-02-05
Import consistency

sebrandon1:
Be more consistent in import naming.  No functional changes.
- `metaV1` to `metav1`.
- `coreV1` to `corev1`.

Similar to: 
- https://github.com/test-network-function/cnf-certification-test/pull/618
- https://github.com/test-network-function/cnfcert-tests-verification/pull/229

PR #270, opened 2024-03-12
Add unit tests for deployment pkg Pull func

sebrandon1:
Builds on top of #261 

Proof of concept PR for adding unit tests wrapping some of the deployment package starting with the `Pull` func.

Notable changes:
- Added func `GetTestClients` which helps build a test client for doing unit testing.
- Changed `settings.K8sClient` to use the Interface instead of the client.
- Changed `settings.NetworkingV1Client` to `settings.NetworkingV1Inteface`.
- Vendor changes come from using the fake k8s client and the additional code needed for that as well as the stretchr/testify library.
- Changes `v1` to `appsv1` similar to #239

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -101,14 +101,20 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 
 	if name == "" {
 		builder.errorMsg = "deployment 'name' cannot be empty"
+
+		return &builder, fmt.Errorf(builder.errorMsg)
 	}
 
 	if nsname == "" {
 		builder.errorMsg = "deployment 'namespace' cannot be empty"
+
+		return &builder, fmt.Errorf(builder.errorMsg)
 	}
 
 	if !builder.Exists() {
-		return nil, fmt.Errorf("deployment object %s doesn't exist in namespace %s", name, nsname)
+		builder.errorMsg = fmt.Sprintf("deployment object %s doesn't exist in namespace %s", name, nsname)
+
+		return &builder, fmt.Errorf(builder.errorMsg)
 	}
 
 	builder.Definition = builder.Object
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -100,15 +100,11 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 	}
 
 	if name == "" {
-		builder.errorMsg = "deployment 'name' cannot be empty"
-
-		return nil, fmt.Errorf(builder.errorMsg)
+		return nil, fmt.Errorf("deployment 'name' cannot be empty")
 	}
 
 	if nsname == "" {
-		builder.errorMsg = "deployment 'namespace' cannot be empty"
-
-		return nil, fmt.Errorf(builder.errorMsg)
+		return nil, fmt.Errorf("deployment 'namespace' cannot be empty")
 	}
 
 	if !builder.Exists() {
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -102,19 +102,17 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 	if name == "" {
 		builder.errorMsg = "deployment 'name' cannot be empty"
 
-		return &builder, fmt.Errorf(builder.errorMsg)
+		return nil, fmt.Errorf(builder.errorMsg)
 	}
 
 	if nsname == "" {
 		builder.errorMsg = "deployment 'namespace' cannot be empty"
 
-		return &builder, fmt.Errorf(builder.errorMsg)
+		return nil, fmt.Errorf(builder.errorMsg)
 	}
 
 	if !builder.Exists() {
-		builder.errorMsg = fmt.Sprintf("deployment object %s doesn't exist in namespace %s", name, nsname)
-
-		return &builder, fmt.Errorf(builder.errorMsg)
+		return nil, fmt.Errorf("deployment object %s doesn't exist in namespace %s", name, nsname)
 	}
 
 	builder.Definition = builder.Object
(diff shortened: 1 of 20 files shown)


sebrandon1:
Once #273 merges I'll rebase this PR on top of the changes there.

pkg/deployment/deployment.go
```diff
@@ -101,14 +101,20 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
...
 		builder.errorMsg = "deployment 'name' cannot be empty"
+
+		return &builder, fmt.Errorf(builder.errorMsg)
 	}
 
 	if nsname == "" {
 		builder.errorMsg = "deployment 'namespace' cannot be empty"
+
+		return &builder, fmt.Errorf(builder.errorMsg)
 	}
```
kononovn, line 113: Please add single verification and remove multiple return above.

	if builder.errorMsg != "" {
		return nil, fmt.Errorf("faield to pull deployment object due to the following error: %s", builder.errorMsg)
	}

sebrandon1: Ah yes I forgot to change this after we talked about the builder errorMsg structure.  I'll change that.

pkg/deployment/deployment.go
```diff
@@ -101,14 +101,20 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
...
 
 	if !builder.Exists() {
-		return nil, fmt.Errorf("deployment object %s doesn't exist in namespace %s", name, nsname)
+		builder.errorMsg = fmt.Sprintf("deployment object %s doesn't exist in namespace %s", name, nsname)
```
kononovn, line 115: I don't think we need to change it. We failed to pull deployment let's return nil instead of invalid builder.

kononovn (changes requested):

pkg/deployment/deployment.go
sebrandon1: Done.

pkg/deployment/deployment.go
sebrandon1: I think the pattern should be if you set an `errorMsg`, you should immediately return.  Let's discuss!

kononovn: ok we can use multiple returns

pkg/deployment/deployment.go
```diff
@@ -101,10 +101,14 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 
 	if name == "" {
 		builder.errorMsg = "deployment 'name' cannot be empty"
+
+		return &builder, fmt.Errorf(builder.errorMsg)
```
kononovn, line 105: Please return nil instead of builder

pkg/deployment/deployment.go
```diff
@@ -101,10 +101,14 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
...
 	}
 
 	if nsname == "" {
 		builder.errorMsg = "deployment 'namespace' cannot be empty"
+
+		return &builder, fmt.Errorf(builder.errorMsg)
```
kononovn, line 111: return &builder, fmt.Errorf("deployment 'namespace' cannot be empty")

kononovn (changes requested):
2 more comments

pkg/deployment/deployment.go
```diff
@@ -101,10 +101,14 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 
 	if name == "" {
 		builder.errorMsg = "deployment 'name' cannot be empty"
+
+		return nil, fmt.Errorf(builder.errorMsg)
 	}
 
 	if nsname == "" {
 		builder.errorMsg = "deployment 'namespace' cannot be empty"
```
kononovn, line 109: This line can be safely removed

pkg/deployment/deployment.go
```diff
@@ -101,10 +101,14 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 
 	if name == "" {
 		builder.errorMsg = "deployment 'name' cannot be empty"
```
kononovn, line 103: This line can be safely removed

kononovn (approved):
/lgtm

pkg/deployment/deployment.go
sebrandon1: Done.

pkg/deployment/deployment.go
sebrandon1: Done.

pkg/deployment/deployment_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/stretchr/testify/assert"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+//nolint:funlen
```
achuzhoy, line 14: Why the exception?

pkg/deployment/deployment.go
sebrandon1: Done.

pkg/deployment/deployment_test.go
```diff
@@ -0,0 +1,114 @@
...
+		if testCase.expectedError {
+			assert.NotNil(t, err)
+
+			// Check the error message
+			if testCase.expectedErrorText != "" {
+				assert.Equal(t, testCase.expectedErrorText, err.Error())
+			}
+		} else {
+			assert.Nil(t, err)
+			assert.Equal(t, testDeployment.Name, builderResult.Object.Name)
```
trewest, line 111: Guess we might as well add an assertion for namespace as well

pkg/deployment/deployment_test.go
sebrandon1: Length of the test function was too long for the linter.

pkg/deployment/deployment_test.go
sebrandon1: Good call I can add that.

sebrandon1: Done.

trewest (approved):
/lgtm

PR #271, opened 2024-03-13
sriov pkg: Added create and newBuilder funcs for sriovoperatorconfig

evgenLevin:


pkg/sriov/operatorconfig.go
```diff
@@ -29,6 +29,56 @@ type OperatorConfigBuilder struct {
 	errorMsg  string
 }
 
+// NewOperatorConfigBuilder creates new instance of OperatorConfigBuilder.
+func NewOperatorConfigBuilder(apiClient *clients.Settings, nsname string) *OperatorConfigBuilder {
+	glog.V(100).Infof(
+		"Initializing new OperatorConfigBuilder structure with the following params: %s", nsname)
```
achuzhoy, line 35: Maybe it's worth to replace param with namespace.
Looking at the log - not clear that the param is nsname

cdvultur: agree with @achuzhoy on this one.

pkg/sriov/operatorconfig.go
```diff
@@ -29,6 +29,56 @@ type OperatorConfigBuilder struct {
...
+		)
+
+		if err != nil {
+			glog.V(100).Infof("Failed to create the SriovOperatorConfig")
+
+			return nil, err
+		}
+	}
+
+	builder.Object = builder.Definition
```
cdvultur, line 77: Why do you overwrite the buidler.Object with builder.Definition ?

evgenLevin: right, no need! removed

cdvultur (approved):
lgtm

kononovn (approved):
/lgtm

PR #303, opened 2024-03-27
pod: modified ExecCommand method's glog Infof

ajaggapa:
Minor modification to ExecCommand method's glog to help during troubleshooting.

Logs to include namespace, pod name and container name in which command is being executed.


kononovn (approved):
/lgtm

PR #308, opened 2024-03-28
sriov pkg: added sriovNetwork unit tests

kononovn:


pkg/sriov/network_test.go
```diff
@@ -0,0 +1,554 @@
...
+
+		var testSettings *clients.Settings
+
+		testNetwork := generateNetwork(testCase.networkName, testCase.networkNamespace)
+
+		if testCase.addToRuntimeObjects {
+			runtimeObjects = append(runtimeObjects, testNetwork)
+		}
+
+		testSettings = nil
```
sebrandon1, line 96: I believe this isn't needed since its just declared as a pointer above.

pkg/sriov/network_test.go
```diff
@@ -0,0 +1,554 @@
...
+			assert.Equal(t, testNetwork.Name, builderResult.Object.Name)
+			assert.Equal(t, testNetwork.Namespace, builderResult.Object.Namespace)
+		}
+	}
+}
+
+func TestNewNetworkBuilder(t *testing.T) {
+	generateNetworkBuilder := NewNetworkBuilder
+
+	var testSettings *clients.Settings
```
sebrandon1, line 124: Move this inside the loop.

pkg/sriov/network_test.go
```diff
@@ -0,0 +1,554 @@
...
+// buildValidTestBuilder returns a valid Builder for testing purposes.
+func buildInvalidSrIovNetworkTestBuilder(apiClient *clients.Settings) *NetworkBuilder {
+	return NewNetworkBuilder(
+		apiClient, defaultNetName, defaultNetNsName, defaultNetTargetNsName, "")
+}
+
+func buildTestClientWithDummyObject() *clients.Settings {
+	runtimeObjects := buildDummySrIovNetworkObject()
+
+	return clients.GetTestClients(runtimeObjects)
```
sebrandon1, line 540:
```suggestion
	return clients.GetTestClients(buildDummySrIovNetworkObject())
```

sebrandon1 (commented):
Some comments!

I think I like the idea of moving `testSettings := buildTestClientWithDummyObject()` inside the testCase loops so its scope is more readable.

achuzhoy (dismissed):

pkg/sriov/network_test.go
kononovn: +1

sebrandon1 (approved):

PR #314, opened 2024-04-02
Add deployment unit tests

sebrandon1:
Resolves: #270 

- Adds unit tests for the deployment package.  
- Modifies the interface used in the `Builder`.

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -124,9 +124,7 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 
 // WithNodeSelector applies a nodeSelector to the deployment definition.
 func (builder *Builder) WithNodeSelector(selector map[string]string) *Builder {
-	if valid, err := builder.validate(); !valid {
-		builder.errorMsg = err.Error()
-
+	if valid, _ := builder.validate(); !valid {
 		return builder
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/deployment/deployment.go
```diff
@@ -118,7 +124,9 @@ func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 
 // WithNodeSelector applies a nodeSelector to the deployment definition.
 func (builder *Builder) WithNodeSelector(selector map[string]string) *Builder {
-	if valid, _ := builder.validate(); !valid {
+	if valid, err := builder.validate(); !valid {
+		builder.errorMsg = err.Error()
```
trewest, line 128: I think it is redundant. builder.errorMsg should already contain what you are registering in err

sebrandon1: I'll double-check what I was trying to do with this one.

pkg/deployment/deployment_test.go
```diff
@@ -431,3 +451,249 @@ func TestWithToleration(t *testing.T) {
...
+
+	for _, testCase := range testCases {
+		var runtimeObjects []runtime.Object
+
+		if testCase.deploymentExistsAlready {
+			runtimeObjects = append(runtimeObjects, generateTestDeployment())
+		}
+
+		testBuilder := buildTestBuilderWithFakeObjects(runtimeObjects)
+		result, err := testBuilder.Update()
```
kononovn, line 519: Can you create test which actually mutate some filed using update function ?

kononovn (changes requested):
Overall looks good. Just 1 comment to fix

pkg/deployment/deployment.go
sebrandon1: I double checked and removed it.  Good catch I'm not sure why I added that but it wasn't needed.  😄

kononovn (approved):
/lgtm

PR #317, opened 2024-04-04
daemonset: IsReady method evaluation modified

ajaggapa:
In cases, when DS cannot be scheduled, IsReady is returning true. Added one more evaluation to ensure reliable result.

_Example:_
```
Desired Number of Nodes Scheduled: 2                                                                                                                                                                              
Current Number of Nodes Scheduled: 0                                                                                                                                                                               
Number of Nodes Scheduled with Up-to-date Pods: 0   
```


kononovn (approved):
/lgtm

PR #318, opened 2024-04-04
sriov: added unit tests for lists functions

kononovn:


pkg/sriov/list_test.go
```diff
@@ -0,0 +1,349 @@
+package sriov
+
+import (
+	"fmt"
+	"testing"
+
+	srIovV1 "github.com/k8snetworkplumbingwg/sriov-network-operator/api/v1"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
sebrandon1, line 10:
```suggestion
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```

sebrandon1 (commented):
One comment about import consistency eg. #239

pkg/sriov/list_test.go
```diff
@@ -0,0 +1,349 @@
...
+	testCases := []struct {
+		testNetwork   []*NetworkBuilder
+		nsName        string
+		listOptions   []v1.ListOptions
+		expectedError error
+		client        bool
+	}{
+		{
+			testNetwork:   []*NetworkBuilder{buildValidSriovNetworkTestBuilder(buildTestClientWithDummyObject())},
+			nsName:        defaultNetNsName,
```
trewest, line 24: IMO, using variable defined in other files makes it hard to understand what you are passing into the test. I think I prefer just using hard coded strings for the test inputs.

sebrandon1: I agree.  I was actually looking for the definition of these vars but couldn't find them 😄

pkg/sriov/list_test.go
```diff
@@ -0,0 +1,349 @@
...
+		{
+			testNetwork:    []*NetworkBuilder{buildValidSriovNetworkTestBuilder(buildTestClientWithDummyObject())},
+			operatorNsName: defaultNetNsName,
+			targetNsName:   defaultNetTargetNsName,
+			client:         true,
+		},
+		{
+			testNetwork:   []*NetworkBuilder{buildValidSriovNetworkTestBuilder(buildTestClientWithDummyObject())},
+			targetNsName:  defaultNetTargetNsName,
+			expectedError: fmt.Errorf("failed to clean up sriov networks, 'operatornsname' parameter is empty"),
```
trewest, line 88: Can you explicitly add operatorNsName: "" so that it is clear why you are expecting the error that you are?

pkg/sriov/list_test.go
```diff
@@ -0,0 +1,349 @@
...
+			client:        true,
+		},
+		{
+			testNetwork:    []*NetworkBuilder{buildValidSriovNetworkTestBuilder(buildTestClientWithDummyObject())},
+			operatorNsName: defaultNetNsName,
+			expectedError:  fmt.Errorf("failed to clean up sriov networks, 'targetnsname' parameter is empty"),
```
trewest, line 94: Can you explicitly add targetNsName: "" so that it is clear why you are expecting the error that you are?

PR #323, opened 2024-04-09
Use client params struct

sebrandon1:
Created a `TestClientParams` struct to house parameters that are needed inside of the `GetTestClients()` func.  This allows for flexibility whenever we need to add new types to our test client.


kononovn (approved):
/lgtm

PR #327, opened 2024-04-12
add new method to the co pkg: Pull function

elenagerman:
clusteroperator: added Pull function

### DIFF
--- a/pkg/clusteroperator/clusteroperator.go
+++ b/pkg/clusteroperator/clusteroperator.go
@@ -35,6 +35,28 @@ type Builder struct {
 	errorMsg string
 }
 
+// Pull loads an existing clusterOperator into Builder struct.
+func Pull(apiClient *clients.Settings, clusterOperatorName string) (*Builder, error) {
+	glog.V(100).Infof("Pulling existing clusterOperator: %s", clusterOperatorName)
+
+	builder := Builder{
+		apiClient: apiClient,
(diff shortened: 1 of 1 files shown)


pkg/clusteroperator/clusteroperator.go
```diff
@@ -145,6 +167,44 @@ func (builder *Builder) WaitUntilConditionTrue(
 		})
 }
 
+// WaitForClusteroperatorAvailability waits until specific clusteroperator will reach available state.
+func WaitForClusteroperatorAvailability(
```
kononovn, line 171: According to code I can see WaitUntilAvailable function already implemented. Can it be used instead of this new function?

You'll need just:
```
coObject, err := clusteroperator.Pull(apiClient, "OPERATOR NAme")
Expect(err).ToNot(HaveOccured())
err := coObject.WaitUntilAvailable()
Expect(err).ToNot(HaveOccured())
```

kononovn (changes requested):
Pull function looks good.
Please check 1 comment above and change commit message to:
clusteroperator: added Pull and WaitForClusteroperatorAvailability funcs

pkg/clusteroperator/clusteroperator.go
elenagerman: sure

kononovn (approved):
/lgtm

PR #342, opened 2024-04-15
ocm: added new placement rule builder and unit tests

klaskosk:
Added `NewPlacementRuleBuilder`  and a unit test for it.

### DIFF
--- a/pkg/ocm/placementrule.go
+++ b/pkg/ocm/placementrule.go
@@ -26,9 +26,46 @@ type PlacementRuleBuilder struct {
 	errorMsg string
 }
 
+// NewPlacementRuleBuilder creates a new instance of PlacementRuleBuilder.
+func NewPlacementRuleBuilder(apiClient *clients.Settings, name, nsname string) *PlacementRuleBuilder {
+	glog.V(100).Infof(
+		"Initializing new placement rule structure with the following params: name: %s, nsname: %s",
+		name, nsname)
+
+	builder := PlacementRuleBuilder{
+		apiClient: apiClient,
+		Definition: &placementrulev1.PlacementRule{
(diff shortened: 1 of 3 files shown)


kononovn (changes requested):
Looks good just few minor comments to fix.

kononovn (approved):
/lgtm

sebrandon1 (commented):
One comment about returns.

pkg/ocm/placementrule.go
```diff
@@ -26,10 +26,47 @@ type PlacementRuleBuilder struct {
...
+		glog.V(100).Info("The name of the PlacementRule is empty")
+
+		builder.errorMsg = "placementrule's 'name' cannot be empty"
+	}
+
+	if nsname == "" {
+		glog.V(100).Info("The namespace of the PlacementRule is empty")
+
+		builder.errorMsg = "placementrule's 'nsname' cannot be empty"
+	}
```
sebrandon1, line 55:
```suggestion
	if name == "" {
		glog.V(100).Info("The name of the PlacementRule is empty")

		builder.errorMsg = "placementrule's 'name' cannot be empty"
		
		return builder
	}

	if nsname == "" {
		glog.V(100).Info("The namespace of the PlacementRule is empty")

		builder.errorMsg = "placementrule's 'nsname' cannot be empty"
		
		return builder
	}
```

klaskosk:
@sebrandon1 Thanks, I made the change to returns

pkg/ocm/placementrule.go
sebrandon1: @klaskosk After speaking with @kononovn we decided your original approach was fine for now.  If you want to revert my above suggestion that would be great.

kononovn (approved):

PR #355, opened 2024-04-18
Address return consistency

sebrandon1:
I went through the entire codebase and addressed inconsistencies with `return` statements to make everything the "same".

Now we can expect:
- All "NewBuilder" funcs to early return if any params are empty.
- All With... and operation funcs are modified to return a nil pointer to the builder and the error (if return values are applicable).


sebrandon1:
Closing as all of the issues with returns have been addressed in other PRs.

PR #358, opened 2024-04-19
ibu: add extra check for stage to avoid going into poll

achuzhoy:


pkg/lca/imagebasedupgrade.go
```diff
@@ -381,6 +384,10 @@ func (builder *ImageBasedUpgradeBuilder) WaitUntilStageComplete(stage string) (*
 		return builder, fmt.Errorf(builder.errorMsg)
 	}
 
+	if stage != idle && stage != "Prep" && stage != "Upgrade" && stage != "Rollback" {
```
trewest, line 387:
```suggestion
	if !slices.Contains([]string{"Idle", "Prep", "Upgrade", "Rollback"}, stage) {
```

achuzhoy: adjusted. Thanks.

trewest (approved):
/lgtm

kononovn (approved):
/lgtm

Issue #361, opened 2024-04-22
Limit api clients and add 100% unit test coverage for ingress pkg


sebrandon1:
Limit API clients for Builder and Pull functions.
Add 100% unit test coverage for the package.

https://github.com/openshift-kni/eco-goinfra/tree/main/pkg/ingress

PR #366, opened 2024-04-25
Add ingress controller unit tests

sebrandon1:
Note: It looks like the `Update()` is fundamentally broken because it does not appear you can modify IngressController objects once they are created.  I tried various things to try and get it to work.

I left my test in there but commented out until we decide what to do with the `Update()` function itself.

Other changes:
* Modified the `Delete()` function to set the object to nil and return to match other package(s) functionality.

Resolves: https://github.com/openshift-kni/eco-goinfra/issues/361

### DIFF
--- a/pkg/ingress/ingress_test.go
+++ b/pkg/ingress/ingress_test.go
@@ -1,1 +1,293 @@
 package ingress
+
+import (
+	"testing"
+
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func TestIngressPull(t *testing.T) {
+	testCases := []struct {
+		ingressName         string
+		ingressNamespace    string
+		expectedError       bool
+		addToRuntimeObjects bool
+		expectedErrorText   string
+	}{
+		{
+			expectedError:       false,
+			addToRuntimeObjects: true,
+			expectedErrorText:   "",
+			ingressName:         "test",
+			ingressNamespace:    "test",
+		},
+		{
+			expectedError:       true,
+			addToRuntimeObjects: false,
+			expectedErrorText:   "ingresscontroller object test not found in namespace test",
+			ingressName:         "test",
+			ingressNamespace:    "test",
+		},
+		{
+			expectedError:       true,
+			addToRuntimeObjects: true,
+			expectedErrorText:   "ingresscontroller object test not found in namespace test",
+			ingressName:         "",
+			ingressNamespace:    "test",
+		},
+		{
+			expectedError:       true,
+			addToRuntimeObjects: true,
+			expectedErrorText:   "ingresscontroller object test not found in namespace test",
+			ingressName:         "test",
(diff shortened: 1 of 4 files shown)


kononovn (changes requested):

pkg/ingress/ingresscontroller.go
```diff
@@ -16,20 +16,22 @@ import (
...
 	// ingresscontroller definition, used to create the ingresscontroller object.
-	Definition *v1.IngressController
+	Definition *operatorv1.IngressController
 	// Created ingresscontroller object.
-	Object *v1.IngressController
-	// api client to interact with the cluster.
-	apiClient *clients.Settings
+	Object *operatorv1.IngressController
+	// api clients to interact with the cluster.
+	readerClient goclient.Reader
```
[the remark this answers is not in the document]
sebrandon1: Done.

pkg/ingress/ingress_test.go
```diff
@@ -1 +1,241 @@
...
+			expectedError:       true,
+			addToRuntimeObjects: false,
+			expectedErrorText:   "ingresscontroller object test not found in namespace test",
+			ingressName:         "test",
+			ingressNamespace:    "test",
+		},
+		{
+			expectedError:       true,
+			addToRuntimeObjects: true,
+			expectedErrorText:   "ingresscontroller object test not found in namespace test",
```
kononovn, line 38: It looks like error is invalid. It has to be "The` ingresscontroller name is empty" no?

pkg/ingress/ingress_test.go
```diff
@@ -1 +1,241 @@
...
+			ingressName:         "",
+			ingressNamespace:    "test",
+		},
+		{
+			expectedError:       true,
+			addToRuntimeObjects: true,
+			expectedErrorText:   "ingresscontroller object test not found in namespace test",
```
kononovn, line 45: expected error has to be different

kononovn (changes requested):
Please check 2 more comments.

pkg/ingress/ingress_test.go
sebrandon1: Fixed.

pkg/ingress/ingress_test.go
sebrandon1: Fixed.

sebrandon1:
Updated for the comments left.  

I removed my commented out test for `Update()` but I'm pretty sure that function doesn't work still.  Do we want to maybe address that in another PR?

pkg/ingress/ingresscontroller_test.go
```diff
@@ -0,0 +1,241 @@
...
+			if testCase.expectedErrorText != "" {
+				assert.Equal(t, testCase.expectedErrorText, err.Error())
+			}
+		} else {
+			assert.Nil(t, err)
+			assert.NotNil(t, builderResult)
+		}
+	}
+}
+
```
kononovn, line 83: Can you add tests for Get(),Exist(), Update - It's OK to have it commented.

kononovn (changes requested):
Please add few more tests.

pkg/ingress/ingresscontroller_test.go
sebrandon1: Done.

kononovn (approved):
/lgtm

PR #368, opened 2024-04-29
WithCustomResourcesRequests method added

yprokule:
Method _WithCustomResourcesRequests_ allows to request custom resources for a container for e.g. _SR-IOV_ resource.

### DIFF
--- a/pkg/pod/container.go
+++ b/pkg/pod/container.go
@@ -280,6 +280,23 @@ func (builder *ContainerBuilder) WithResourceRequest(hugePages, memory string, c
 	return builder
 }
 
+// WithCustomResourcesRequests applies custom resource requests struct on container.
+func (builder *ContainerBuilder) WithCustomResourcesRequests(resourceList corev1.ResourceList) *ContainerBuilder {
+	glog.V(100).Infof("Applying custom resource requests to container: %v", resourceList)
+
+	if len(resourceList) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/pod/container.go
```diff
@@ -280,6 +280,25 @@ func (builder *ContainerBuilder) WithResourceRequest(hugePages, memory string, c
...
+func (builder *ContainerBuilder) WithCustomResourcesRequests(resourceList corev1.ResourceList) *ContainerBuilder {
+	glog.V(100).Infof("Applying custom resource requests to container: %v", resourceList)
+
+	if len(resourceList) == 0 {
+		glog.V(100).Infof("Container's resource limit var 'resourceList' is empty")
+
+		builder.errorMsg = "container's resource limit var 'resourceList' is empty"
+	}
+
+	if builder.errorMsg != "" {
```
cdvultur, line 293: I think this can be updated based on https://github.com/openshift-kni/eco-goinfra/pull/355/files 

If there is already an error, we can return directly the builder in the previous if block.

cdvultur (commented):
small comment on return consistency

pkg/pod/container.go
yprokule: thanks, updated

achuzhoy (approved):

trewest (approved):
/lgtm

PR #379, opened 2024-04-30
Delete() funcs should return nil if not exists

sebrandon1:
There seems to be a mismatch of how we expect `Delete()` funcs throughout the repo.

The general consensus from most of the `Delete()` functions is that we should return nil because the object doesn't exist.  There are a few outliers that I changed in this PR that were returning an error because an item didn't exist.

I have adjusted all of the Delete() funcs to set the `builder.Object` to nil and then return `nil`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -83,6 +83,8 @@ linters-settings:
     checks:
       - all
       - ST1001
+      - "-SA1006"
+    
 
 linters:
   enable:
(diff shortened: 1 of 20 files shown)


sebrandon1:
Closing!  All of the necessary changes have been broken down and merged in via their own respective Pull Requests.  Thanks for all of the reviews and merges!

PR #400, opened 2024-05-07
lca: updating the seedgenerator - singleton

achuzhoy:
The seedgenerator can have a specific name only.
Updating respectively.


kononovn (approved):
/lgtm

PR #402, opened 2024-05-08
cgu: added delete and wait to force update

klaskosk:
This fixes an issue where force updates could create the cgu before it was deleted, resulting in it just being deleted.


kononovn (approved):
/lgtm

PR #404, opened 2024-05-08
nodes and apiservers pkgs: reduce glog level from 100 to 80

elenagerman:
the default is 100 for almost all our methods and it's very hard to debug on nodes reboots

### DIFF
--- a/pkg/apiservers/kubeapiserver.go
+++ b/pkg/apiservers/kubeapiserver.go
@@ -101,7 +101,7 @@ func (builder *KubeAPIServerBuilder) GetCondition(conditionType string) (*operat
 		return nil, "", err
 	}
 
-	glog.V(100).Infof("Get %s kubeAPIServer %s condition", builder.Definition.Name, conditionType)
+	glog.V(80).Infof("Get %s kubeAPIServer %s condition", builder.Definition.Name, conditionType)
 
 	if conditionType == "" {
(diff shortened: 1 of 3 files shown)


elenagerman:
@kononovn @trewest @achuzhoy WDYT?

kononovn:
@elenagerman According to our original agreement all logs in eco-goinfra project should have level 100. If you want to reduce number of logs your could delete them.

kononovn:
@elenagerman If you want to hide all logs from eco-goinfra you can set env var export ECO_VERBOSE_LEVEL below 100

elenagerman:
> @elenagerman If you want to hide all logs from eco-goinfra you can set env var export ECO_VERBOSE_LEVEL below 100

I didn't planed to hide all logs. The problem is that during long wait-for-something methods you getting such logs non stop and, on some stage losing relevant logs.
As an example let's take  WaitForAllNodesToReboot method:
 Wait for the reboot 1 node takes at least 5 mins. according to the pool size it will be  biggest pool size * 5 mins.
 Each 2 second both Exists and IsReady print 1 message.
Maybe we need to consider this option and to make some role related for glog??

kononovn:
@elenagerman There are 2 options
1. Redirect logs to file then you'll not loose any logs 
2. Delete these logs from functions.
Personally I would prefer first option.

PR #409, opened 2024-05-10
cgu: fix bug in wait until deleted function

klaskosk:
In #402 I accidentally used an update instead of a get inside the WaitUntilDeleted function that results in an error if the cgu is still present after the first check. This is the same situation that led to an error before albeit for a different reason.


kononovn (approved):
/lgtm

PR #340, opened 2024-04-15
Add ManagedCluster to olm package

ddmitrie:

PR #416, opened 2024-05-13
ocm: add ManagedCluster and unit tests

klaskosk:
Updated #340 to include unit tests and move back to go 1.20 for now


pkg/ocm/managedcluster.go
```diff
@@ -0,0 +1,201 @@
...
+
+// ManagedClusterAdditionalOptions additional options for ManagedCluster object.
+type ManagedClusterAdditionalOptions func(builder *ManagedClusterBuilder) (*ManagedClusterBuilder, error)
+
+// NewManagedClusterBuilder creates a new instance of ManagedClusterBuilder.
+func NewManagedClusterBuilder(apiClient *clients.Settings, name string) *ManagedClusterBuilder {
+	glog.V(100).Infof(
+		"Initializing new ManagedCluster structure with the following params: name: %s", name)
+
+	builder := &ManagedClusterBuilder{
```
trewest, line 33: Please check if apiClient is nil and return nil if so

trewest (changes requested):
Couple comments

pkg/ocm/managedcluster.go
klaskosk: I don't think any of the other packages return early on a nil apiClient in NewXXBuilder(), not sure if that should be made a global change similar to #355. I've changed it here though

klaskosk:
@trewest Thanks for the comments, I updated the PR

pkg/ocm/managedcluster.go
trewest: It's a requirement for any resources that have had their apiClient limited ([example](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/lca/seedgenerator.go#L43-L47)). Otherwise, you can cause a panic on line 40

trewest (approved):
/lgtm

pkg/ocm/managedcluster.go
klaskosk: Ah makes sense, I'll go ahead and update the cgu package then since I think it'd apply there too

kononovn (approved):
/lgtm

PR #420, opened 2024-05-14
bmh: fixed occasional failed unit test

klaskosk:
In `baremetalhost_test.go` the `TestBareMetalHostWaitUntilDeleted` unit test waits only 1 second for deletion, which causes tests to fail sometimes since the actual WaitUntilDeleted function polls every 1 second and does not immediately execute. This fix just increases the wait time in the test to 2 seconds which should always be sufficient.


kononovn (approved):
/lgtm

PR #424, opened 2024-05-14
BMC: Downgrade gofish dependency to v0.15.0

greyerof:
Temporary downgrade of gofish dependency to avoid mismatch with current one in eco-gotests.


kononovn (approved):
/lgtm

PR #419, opened 2024-05-14
BMC: Added CLI commands runner and Serial Console access.

greyerof:
BMC's CLI commands can be run using the `bmc.RunCLICommand` method, which connects to the CLI via SSH. The method blocks until the command ends or a timeout happens, whatever comes first. The command's output is copied into the stdout & stderr string vars. The underlying SSH session is created and closed internally after the command has finished.

The new method `bmc.OpenSerialConsole` returns a piped reader and writer interfaces that can be used to interactively receive/send commands/text to/from the serial console. Since the serial console is tunneled in a SSH session, the `bmc.CloseSerialConsole` method must be called to release that session to avoid leaks that would prevent CLI commands to fail due to maximum concurrent SSH sessions reached in the BMC.

Also, added SSH port as parameter for BMC's New() method

PR #425, opened 2024-05-14
bmc: add power on/off and power usage

klaskosk:
Might be worth waiting till #419 is merged before merging this PR.

### DIFF
--- a/pkg/bmc/bmc.go
+++ b/pkg/bmc/bmc.go
@@ -53,11 +53,12 @@ type TimeOuts struct {
 
 // BMC is the holder struct for BMC access through redfish & ssh.
 type BMC struct {
-	host        string
-	redfishUser User
-	sshUser     User
-	sshPort     uint16
-	systemIndex int
+	host              string
+	redfishUser       User
+	sshUser           User
+	sshPort           uint16
+	systemIndex       int
+	powerControlIndex int
 
 	timeOuts TimeOuts
 
@@ -121,12 +122,13 @@ func New(host string, redfishUser, sshUser User, sshPort uint16, timeOuts TimeOu
 	}
 
 	return &BMC{
-		host:        host,
-		redfishUser: redfishUser,
-		sshUser:     sshUser,
-		sshPort:     sshPort,
-		timeOuts:    timeOuts,
-		systemIndex: 0,
+		host:              host,
+		redfishUser:       redfishUser,
+		sshUser:           sshUser,
+		sshPort:           sshPort,
+		timeOuts:          timeOuts,
+		systemIndex:       0,
+		powerControlIndex: 0,
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/bmc/bmc.go
```diff
@@ -347,3 +435,32 @@ func redfishGetSystemSecureBoot(redfishClient *gofish.APIClient, systemIndex int
...
+		return nil, fmt.Errorf("failed to get chassis collection: %w", err)
+	}
+
+	for chassisIndex, chassis := range chassisCollection {
+		power, err := chassis.Power()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get power for chassis index %d: %w", chassisIndex, err)
+		}
+
+		if power == nil {
```
greyerof, line 453: I'm curious: is this just overly defensive programming or is it possible having nil pointers for some of the elements of this slice?

yprokule:
@klaskosk could U please also add support for `power cycle`?

pkg/bmc/bmc.go
klaskosk: Yes, `chassis.Power()` returns `nil, nil` if there is no link to power for that chassis, I already got hit with a nil pointer error for that :)

Ideally you should be able to specify which chassis to get the power from, but we can't use an index since I found out the hard way that gofish fetches all the chassis concurrently so they're not in the same order the api returns them.

This is just the method I went with since the only chassis with a power link is the one that I want for my use cases, but it's not as robust as I'd like

klaskosk:
@yprokule Sounds good, I'll add that today!

pkg/bmc/bmc.go
greyerof: I see, ok. So I didn't know that about the concurrent requests... gofish uses the same mechanism for every collection, so using that index is probably a bad idea :(

klaskosk: For `PowerControl` in this PR it'll be fine since they're not links, but yes we probably need a better way to identify systems. Maybe we store an optional `systemId` and default to just using the first system returned otherwise? I expect most of my use cases to be only one system per BMC but good to leave the option for multiple systems open

klaskosk: The checking for len of power.PowerControl is just making sure the index is not out of bounds. I kept the form from `redfishGetSystem` for symmetry but I think it would be clearer to write `powerControlIndex >= len(power.PowerControl)` as the condition.

But the iteration is to check for the first chassis that has the power link, since it there may be multiple enclosures that don't have the power link. It's definitely not ideal though, since I really want the chassis that has the same Id as the system Id

greyerof: I'm ok with that as a workaround, but we should definitely consider/explore other options to improve this in further PR/s. I can investigate this.

klaskosk: Thanks for investigating, I'm open to changing this in the future!

greyerof (approved):

pkg/bmc/bmc_test.go
```diff
@@ -40,6 +57,8 @@ type redfishAPIResponseCallbacks struct {
 	sessions   func(r *http.Request)
 	system     func(r *http.Request)
 	secureBoot func(r *http.Request)
+	chassis    func(r *http.Request)
+	power      func(r *http.Request)
 }
 
 func getDelayResponseCallbackFn(t *testing.T, respDelay time.Duration) func(r *http.Request) {
```
kononovn, line 64: I missed that. Can you please move all private function to the end of the file.

pkg/bmc/bmc_test.go
```diff
@@ -305,6 +351,43 @@ func TestBMC_SetSystemIndex(t *testing.T) {
 	}
 }
 
+func TestBMC_SetPowerControlIndex(t *testing.T) {
```
kononovn, line 354: Can you please remove underscore from all public test functions?

kononovn (changes requested):
Minor change required.

klaskosk:
@kononovn Thanks for the review, I've made those changes

PR #426, opened 2024-05-14
bmh: fix test for delete and wait

klaskosk:
Same issue as in #420, the `DeleteAndWaitUntilDeleted` test will occasionally fail. Didn't notice this function needed to be changed during that PR.


kononovn (approved):
/lgtm

PR #427, opened 2024-05-15
Add MultiClusterEngine to OCM package

ddmitrie:


achuzhoy:
is OCM the right place?
cc @trewest @vkolodny

trewest:
@ddmitrie All new public functions/methods require unit tests. Can you add unit tests following these guidelines? https://github.com/openshift-kni/eco-goinfra?tab=readme-ov-file#note-every-new-package-requires-a-coverage-of-all-its-public-functions-with-unit-tests-unit-tests-are-located-in-the-same-package-as-the-resource-in-a-file-with-the-name-resource_testgo-examples-can-be-found-in-configmap_testgo-and-metallb_testgo

pkg/ocm/multiclusterengine.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+
+	return builder
+}
+
+// PullMultiClusterEngine loads an existing MultiClusterEngine into MultiClusterEngineBuilder struct.
+func PullMultiClusterEngine(apiClient *clients.Settings, name string) (*MultiClusterEngineBuilder, error) {
+	glog.V(100).Infof("Pulling existing MultiClusterEngine name: %s", name)
+
```
trewest, line 81: Please check that apiClient is not nil before accessing it's members

pkg/ocm/multiclusterengine.go
```diff
@@ -0,0 +1,207 @@
...
+// Get fetches the defined MultiClusterEngine from the cluster.
+func (builder *MultiClusterEngineBuilder) Get() (*mceV1.MultiClusterEngine, error) {
+	if valid, err := builder.validate(); !valid {
+		return nil, err
+	}
+
+	glog.V(100).Infof("Getting MultiClusterEngine %s", builder.Definition.Name)
+
+	MultiClusterEngine := &mceV1.MultiClusterEngine{}
```
trewest, line 112:
```suggestion
	multiclusterengine := &mceV1.MultiClusterEngine{}
```

pkg/ocm/multiclusterengine.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// Update modifies an existing MultiClusterEngine on the cluster.
+func (builder *MultiClusterEngineBuilder) Update() (*MultiClusterEngineBuilder, error) {
+	if valid, err := builder.validate(); !valid {
+		return builder, err
+	}
+
+	glog.V(100).Infof("Updating MultiClusterEngine %s", builder.Definition.Name)
+
```
trewest, line 131: Please check that the multiclusterengine exists before attempting to update

pkg/ocm/multiclusterengine.go
```diff
@@ -0,0 +1,207 @@
...
+	err := builder.apiClient.Update(context.TODO(), builder.Definition)
+	builder.Object = builder.Definition
```
trewest, line 133: This runs regardless of if the above operation is successful or not. if err != nil, you should return without updating builder.Object

pkg/ocm/multiclusterengine.go
```diff
@@ -0,0 +1,212 @@
...
+// Delete removes a MultiClusterEngine from the cluster.
+func (builder *MultiClusterEngineBuilder) Delete() error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof("Deleting the MultiClusterEngine %s", builder.Definition.Name)
+
+	if !builder.Exists() {
+		return fmt.Errorf("multiclusterengine cannot be deleted because it does not exist")
```
yprokule, line 152: if the object doesn't exist and _Delete_ method is invoked probably worth returning success?

trewest:
Converted to Draft. Seems like there are some conflicts between this pkg and the assisted pkg. We need to implement these APIs ourselves to address them.

PR #429, opened 2024-05-16
sriov pkg: Renamed CleanAllPoolConfigs func

evgenLevin:
Design was changed: there is no default sriovpoolconfig.


kononovn (approved):
/lgtm

PR #434, opened 2024-05-21
Add unit tests for agentclusterinstall

trewest:
Adds unit tests for the agentclusterinstall resource in the assisted pkg

Needed for https://github.com/openshift-kni/eco-goinfra/issues/295

### DIFF
--- a/pkg/assisted/agentclusterinstall_test.go
+++ b/pkg/assisted/agentclusterinstall_test.go
@@ -1,0 +1,1191 @@
+package assisted
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	hiveextV1Beta1 "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
+	v1 "github.com/openshift/hive/apis/hive/v1"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	fakeRuntimeClient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+const (
+	aciTestName      = "aci-test-name"
+	aciTestNamespace = "aci-test-namespace"
+)
+
+func TestNewAgentClusterInstallBuilder(t *testing.T) {
+	testCases := []struct {
+		name              string
+		namespace         string
+		clusterDeployment string
+		masterCount       int
+		workerCount       int
+		network           hiveextV1Beta1.Networking
+		client            bool
+		expectedError     string
+	}{
+		{
+			name:              aciTestName,
+			namespace:         aciTestNamespace,
+			clusterDeployment: "aci-test-clusterdeployment",
+			masterCount:       3,
+			workerCount:       2,
+			network:           dummyTestNetwork(),
+			client:            true,
+			expectedError:     "",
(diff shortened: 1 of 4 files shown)


pkg/assisted/agentclusterinstall_test.go
```diff
@@ -0,0 +1,1180 @@
...
+		{
+			count:         1,
+			expectedError: "",
+		},
+		{
+			count:         0,
+			expectedError: "agentclusterinstall controlplane agents must be greater than 0",
+		},
+		{
+			count:         -1,
```
achuzhoy, line 606: isn't this one redundant?

pkg/assisted/agentclusterinstall_test.go
```diff
@@ -0,0 +1,1180 @@
...
+
+func TestWithImageSet(t *testing.T) {
+	testCases := []struct {
+		image         string
+		expectedError string
+	}{
+		{
+			image:         "4.16",
+			expectedError: "",
+		},
```
achuzhoy, line 628: One more test with something that raises an error?

pkg/assisted/agentclusterinstall_test.go
```diff
@@ -0,0 +1,1180 @@
...
+		{
+			count:         1,
+			expectedError: "",
+		},
+		{
+			count:         0,
+			expectedError: "",
+		},
+		{
```
achuzhoy, line 657: seems redundant

achuzhoy (approved):

pkg/assisted/agentclusterinstall_test.go
trewest: 0 and -1 are used to catch edged cases. If WithControlPlaneAgents has it's validation for agentCount removed or changed to allow 0, these tests will catch it.

pkg/assisted/agentclusterinstall_test.go
trewest: Currently this method doesn't return any errors. I will see if there is a validation that can be added before. It was allowed to be empty in the past

pkg/assisted/agentclusterinstall_test.go
trewest: Which part?

pkg/assisted/agentclusterinstall_test.go
trewest: No, it looks like an empty imageset reference can still be passed in

pkg/assisted/agentclusterinstall_test.go
```diff
@@ -0,0 +1,1179 @@
...
+				builder.Definition.Spec.HoldInstallation = true
+
+				return builder, nil
+			},
+			validator: func(acib *AgentClusterInstallBuilder) bool {
+				return acib.Definition.Spec.HoldInstallation == true
+			},
+		},
+		{
+			option: func(builder *AgentClusterInstallBuilder) (*AgentClusterInstallBuilder, error) {
```
kononovn, line 402: Add option which returns simple error.

pkg/assisted/agentclusterinstall_test.go
```diff
@@ -0,0 +1,1179 @@
...
+
+		_, err := testCase.option(testBuilder)
+
+		assert.Nil(t, err)
+
+		assert.True(t, testCase.validator(testBuilder))
+	}
+}
+
+func TestWithApiVip(t *testing.T) {
```
kononovn, line 427: Please rename all functions below to:
TestAgentClusterInstallWith*

kononovn (changes requested):
2 minor comments

kononovn (approved):
/lgtm

PR #448, opened 2024-05-31
keda pkg: kedaController object handler added

elenagerman:


go.mod
```diff
@@ -28,27 +28,29 @@ require (
...
+	k8s.io/utils v0.0.0-20240102154912-e7106e64919e
 	maistra.io/api v0.0.0-20230704084350-dfc96815fb16
 	open-cluster-management.io/governance-policy-propagator v0.12.0
 	open-cluster-management.io/multicloud-operators-subscription v0.11.0
-	sigs.k8s.io/controller-runtime v0.16.3
+	sigs.k8s.io/controller-runtime v0.17.1
 )
 
 require (
+	github.com/kedacore/keda-olm-operator v0.0.0-20240501182040-762f6be5a942
```
kononovn, line 49: Please use v2.12.0 in order to avoid dependency conflict.

kononovn (changes requested):
Please use another version of operator

go.mod
elenagerman: go: errors parsing go.mod:
go.mod:51:2: require github.com/kedacore/keda-olm-operator: version "v2.12.0" invalid: should be v0 or v1, not v2

elenagerman: it is not "github.com/kedacore/keda/v2/apis/keda/v1alpha1",
but ""github.com/kedacore/keda-olm-operator/apis/keda/v1alpha1""

elenagerman: it's a different repo

elenagerman:
> Please use another version of operator

@kononovn there is no v2.12.0 for the ""github.com/kedacore/keda-olm-operator/apis/keda/v1alpha1""

you are talking about different repo   "github.com/kedacore/keda/v2/apis/keda/v1alpha1".
This second one is using a correct version,but it's another MR: https://github.com/openshift-kni/eco-goinfra/pull/449

pkg/keda/kedacontroller.go
```diff
@@ -0,0 +1,310 @@
...
+	kedav1alpha1 "github.com/kedacore/keda-olm-operator/apis/keda/v1alpha1"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// KedaControllerBuilder provides a struct for KedaController object from the cluster and a KedaController definition.
+type KedaControllerBuilder struct {
```
kononovn, line 17: You have "keda" in package name. Please remove keda from structs and functions names.
For example:

- type ControllerBuilder
- func NewControllerBuilder
- func PullKController
...

Same for test function names

kononovn (changes requested):
Overall looks good. Jew few minor changes are needed.

pkg/keda/kedacontroller.go
elenagerman: same package keda will include scaleObject and triggerAuthentication and maybe more objects will be neede, I think it will be better to leave like this (with a full obj name) 
this is api object name = "KedaController"

elenagerman: @kononovn there are many different types of object related to the keda operator and all have a different name, not always starts from "keda" 
and keda is an operator name

kononovn: @elenagerman It's go best practice. 
The point is.
Indead of 
kede. NewKedaControllerBuilder

simply use:
keda. NewControllerBuilder

As a result we don't use keda multiple times and at the same time we undertand that Controller is related to keda only.

kononovn (changes requested):
Looks good just last comment to fix: https://github.com/openshift-kni/eco-goinfra/pull/448/files#r1624615218

elenagerman:
@kononovn now it start to fail on the integration stage, I have no idea what is a reason

elenagerman:
@kononovn I don't understand what is standing behind an idea to add integration with eco-gotest to the MR check.
Now we'll need to open two MRs? Anyway I do not understand how it should work? Code that was not merged yet in eco-goinfra need to be synced with eco-gotests?

kononovn (approved):
/lgtm

cdvultur (approved):
lgtm

PR #450, opened 2024-05-31
keda pkg: ScaleObject object handler added

elenagerman:
ScaleObject object handler added to the keda pkg

### DIFF
--- a/pkg/keda/scaleobject.go
+++ b/pkg/keda/scaleobject.go
@@ -14,12 +14,12 @@ import (
 	goclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ScaleObjectBuilder provides a struct for ScaledObject object from the cluster
+// ScaledObjectBuilder provides a struct for ScaledObject object from the cluster
 // and a ScaledObject definition.
-type ScaleObjectBuilder struct {
+type ScaledObjectBuilder struct {
 	// ScaledObject definition, used to create the ScaledObject object.
 	Definition *kedav2v1alpha1.ScaledObject
-	// Created TriggerAuthentication object.
+	// Created ScaledObject object.
 	Object *kedav2v1alpha1.ScaledObject
 	// Used to store latest error message upon defining or mutating ScaledObject definition.
 	errorMsg string
@@ -27,9 +27,9 @@ type ScaleObjectBuilder struct {
 	apiClient goclient.Client
 }
 
-// NewScaledObjectBuilder creates a new instance of ScaleObjectBuilder.
+// NewScaledObjectBuilder creates a new instance of ScaledObjectBuilder.
 func NewScaledObjectBuilder(
-	apiClient *clients.Settings, name, nsname string) *ScaleObjectBuilder {
+	apiClient *clients.Settings, name, nsname string) *ScaledObjectBuilder {
 	glog.V(100).Infof(
 		"Initializing new scaledObject structure with the following params: "+
 			"name: %s, namespace: %s", name, nsname)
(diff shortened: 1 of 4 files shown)


pkg/nodes/node.go
```diff
@@ -163,8 +163,6 @@ func (builder *Builder) Exists() bool {
 		return false
 	}
 
-	glog.V(100).Infof("Checking if node %s exists", builder.Definition.Name)
```
yprokule, line 166: don't think this belongs here

elenagerman:
@trewest @kononovn @achuzhoy @cdvultur @yprokule 
PTAL

kononovn (approved):
1 cosmetic comment to fix.
/lgtm

pkg/keda/scaleobject_test.go
```diff
@@ -0,0 +1,593 @@
...
+			name:          defaultScaledObjectName,
+			namespace:     "",
+			expectedError: "scaledObject 'nsname' cannot be empty",
+			client:        true,
+		},
+		{
+			name:          defaultScaledObjectName,
+			namespace:     defaultScaledObjectNamespace,
+			expectedError: "",
+			client:        false,
```
cdvultur, line 140: If there should be an error if the client is `false`. In this case the builder needs a check the client is being passed.

pkg/keda/scaleobject_test.go
```diff
@@ -0,0 +1,593 @@
...
+			assert.NotNil(t, result)
+			assert.Equal(t, testCase.testCooldownPeriod, *result.Definition.Spec.CooldownPeriod)
+		}
+	}
+}
+
+func TestScaleObjectWithMinReplicaCount(t *testing.T) {
+	testCases := []struct {
+		testMinReplicaCount int32
+		expectedError       bool
```
cdvultur, line 422: Do you plan to add more testcases ? Otherwise only one of the 2 `expectedError` / `expectedErrorText` could be used

pkg/keda/scaleobject_test.go
```diff
@@ -0,0 +1,593 @@
...
+			assert.NotNil(t, result)
+			assert.Equal(t, testCase.testMinReplicaCount, *result.Definition.Spec.MinReplicaCount)
+		}
+	}
+}
+
+func TestScaleObjectWithMaxReplicaCount(t *testing.T) {
+	testCases := []struct {
+		testMaxReplicaCount int32
+		expectedError       bool
```
cdvultur, line 456: Same here, related to `expectedError`/`expectedErrorText`

cdvultur (changes requested):
It would be nice to move all references to `scaled` , for consistency. 
Also the apiClient should be check if is empty.

pkg/keda/scaleobject.go
```diff
@@ -0,0 +1,358 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// ScaleObjectBuilder provides a struct for ScaledObject object from the cluster
+// and a ScaledObject definition.
+type ScaleObjectBuilder struct {
+	// ScaledObject definition, used to create the ScaledObject object.
+	Definition *kedav2v1alpha1.ScaledObject
+	// Created TriggerAuthentication object.
```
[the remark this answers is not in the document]
elenagerman: yep, thanks!

pkg/keda/scaleobject_test.go
elenagerman: there is no object to get an error from it

elenagerman:
> It would be nice to move all references to `scaled` , for consistency. Also the apiClient should be check if is empty.

client:        false option == empty apiClient. I don't see any other way to checkit

PR #454, opened 2024-06-03
BMC: PowerCycle alternate implementation

greyerof:
Refactored the Power Cycle system reset to perform a PowerOff + On reset actions as a fallback mechanism when PowerCycle reset type is not supported by the system.

Also fixes in some traces.


kononovn (approved):
/lgtm

klaskosk (approved):
/lgtm

PR #455, opened 2024-06-03
Add eco-gotests 'make vet' to PR checks

sebrandon1:
Upon new PRs, clone both repos down (eco-gotests and eco-goinfra) and then replace the eco-goinfra with the local copy in the eco-gotests go.mod.  Once the replacement is done, run the `make vet` path in the eco-gotests folder to see if everything compiles correctly.

This was previously not possible to test because `eco-gotests` was private.

image

### DIFF
--- a/.github/workflows/eco-gotests-integration.yml
+++ b/.github/workflows/eco-gotests-integration.yml
@@ -17,14 +17,17 @@ jobs:
 
     steps:
       - name: Set up Go 1.22
+        if: ${{ !contains(github.event.*.labels.*.name, 'ignore-dep-check') }}
         uses: actions/setup-go@v5
         with:
           go-version: 1.22.3
 
       - name: Check out the eco-goinfra code
+        if: ${{ !contains(github.event.*.labels.*.name, 'ignore-dep-check') }}
         uses: actions/checkout@v4
 
       - name: Check out the eco-gotests code
+        if: ${{ !contains(github.event.*.labels.*.name, 'ignore-dep-check') }}
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/eco-gotests-integration.yml
```diff
@@ -0,0 +1,43 @@
+name: eco-gotests integration
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - main
```
kononovn, line 7: We want to run similar check for releases as well. 
For instance:
We sync eco-gotests release-4.15 with eco-goinfra release-4.15

.github/workflows/eco-gotests-integration.yml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
```
kononovn, line 11: Can we use ubunutu-latest instead?

sebrandon1: I like to specify the exact version.  Recently they made `ubuntu-24.04` available and within a couple of days those runners had random timeout failures that didn't affect the 22.04 runners.  At least when you specify you know exactly what environment you're getting.

achuzhoy (dismissed):

cdvultur (approved):
/lgtm

sebrandon1:
If the `ignore-dep-check` label is set, the whole action is skipped.

image

This will be useful when incompatibilies are not avoidable.

kononovn (approved):
/lgtm

PR #461, opened 2024-06-05
Add 'make coverage-html' make path

sebrandon1:
`make coverage-html` generates a coverage output website.  `*.out` is already covered in the gitignore.


kononovn (approved):
/lgtm

PR #463, opened 2024-06-05
Add Validate tests for pkg/cgu

sebrandon1:
These two files were missing tests for their versions of `Validate()`.


kononovn (approved):
/lgtm

PR #466, opened 2024-06-06
storage: add pull and deletes plus tests to storageclass

klaskosk:
Added a PullClass function to the storage package; added Delete, DeleteAndWait, and WaitUntilDeleted methods to ClassBuilder; and added unit tests for all of the functions in `storageclass.go`.

### DIFF
--- a/pkg/storage/storageclass.go
+++ b/pkg/storage/storageclass.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
@@ -11,6 +12,7 @@ import (
 	storageV1 "k8s.io/api/storage/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
 
 // ClassBuilder provides struct for storageclass object containing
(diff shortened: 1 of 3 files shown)


kononovn (changes requested):
Looks good just few changes are needed.

kononovn:
> @kononovn Updated the test, although there's no check for a nil client in `NewClassBuilder` since the `apiClient` field hasn't been narrowed so it would be redundant with `validate`

I know just to make it more consistent can you add to NewBuilder following condition?
https://github.com/openshift-kni/eco-goinfra/blob/30c5d20d5b39439e9d23a9740f2b5f6eef4ac9ab/pkg/assisted/agentclusterinstall.go#L51

kononovn (changes requested):
please check comment above.

kononovn (dismissed):
/lgtm

pkg/storage/storageclass.go
```diff
@@ -217,6 +261,52 @@ func (builder *ClassBuilder) Delete() error {
...
 }
 
+// DeleteAndWait deletes the StorageClass and waits up to timeout until it has been removed.
+func (builder *ClassBuilder) DeleteAndWait(timeout time.Duration) error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof(
+		"Deleting StorageClass %s and waiting up to %s until it is removed", timeout, builder.Definition.Name)
```
trewest, line 271: Looks like the variables are ordered incorrectly

trewest (approved):
/lgtm

PR #474, opened 2024-06-12
cgu: add wait for condition and unit tests

klaskosk:
In TALM and ZTP test cases there are many scenarios where we are waiting for a CGU to be in some condition. WaitForCondition generalizes these cases (and the existing WaitUntilComplete) into a single function. Unit tests have been added for WaitForCondition as well as the existing WaitUntilComplete.

### DIFF
--- a/pkg/cgu/cgu.go
+++ b/pkg/cgu/cgu.go
@@ -3,6 +3,7 @@ package cgu
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -15,10 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-const (
-	isTrue     = "True"
-	isComplete = "Succeeded"
-)
+var conditionComplete = metav1.Condition{Type: "Succeeded", Status: metav1.ConditionTrue}
 
 // CguBuilder provides struct for the cgu object containing connection to
 // the cluster and the cgu definitions.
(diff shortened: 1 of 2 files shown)


pkg/cgu/cgu.go
```diff
@@ -338,79 +336,55 @@ func (builder *CguBuilder) WaitUntilDeleted(timeout time.Duration) error {
...
-
-// WaitUntilComplete waits the specified timeout for the CGU to complete.
-func (builder *CguBuilder) WaitUntilComplete(timeout time.Duration) (*CguBuilder, error) {
+// WaitForCondition waits until the CGU has a condition that matches the expected, checking only the Type, Status,
+// Reason, and Message fields. For the message field, it matches if the message contains the expected. Zero fields in
+// the expected condition are ignored.
+func (builder *CguBuilder) WaitForCondition(expected metav1.Condition, timeout time.Duration) (*CguBuilder, error) {
 	if valid, err := builder.validate(); !valid {
-		return builder, err
+		return nil, err
```
cdvultur, line 344: From what I see, only in `Get()` functions we return the nil builder. Wonder why you want to do that in a `Wait()` function

cdvultur (commented):
Looks good. Only one question to clarify

sebrandon1 (approved):

pkg/cgu/cgu.go
klaskosk: @cdvultur I think changing to return nil and an error is one of the goals of #355

kononovn: Please revert it back to:
`return builder, err
`

pkg/cgu/cgu.go
```diff
@@ -338,79 +336,55 @@ func (builder *CguBuilder) WaitUntilDeleted(timeout time.Duration) error {
...
 	}
 
-	glog.V(100).Infof("Waiting for CGU %s to complete", builder.Definition.Name)
-
 	if !builder.Exists() {
 		glog.V(100).Infof("The CGU does not exist on the cluster")
 
-		return builder, fmt.Errorf(builder.errorMsg)
+		return nil, fmt.Errorf(
```
kononovn, line 350: Please revert it back to:
`return builder, fmt.Errorf`

pkg/cgu/cgu.go
```diff
@@ -338,79 +336,55 @@ func (builder *CguBuilder) WaitUntilDeleted(timeout time.Duration) error {
...
+		context.TODO(), 3*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+			var err error
+			builder.Object, err = builder.apiClient.RanV1alpha1().ClusterGroupUpgrades(builder.Definition.Namespace).
+				Get(context.TODO(), builder.Definition.Name, metav1.GetOptions{})
 
 			if err != nil {
-				return false, nil
+				glog.V(100).Info("failed to get cgu %s/%s: %w", builder.Definition.Name, builder.Definition.Namespace, err)
+
+				return false, err
```
kononovn, line 363: If you return error in  wait.PollUntilContextTimeout it'll cause code failure. I think it's not the goal.
We don't care about any errors or issues except timeout. Please revert it back to nil.

pkg/cgu/cgu.go
```diff
@@ -338,79 +336,55 @@ func (builder *CguBuilder) WaitUntilDeleted(timeout time.Duration) error {
...
 			}
 
 			builder.Definition = builder.Object
 
 			for _, condition := range builder.Object.Status.Conditions {
-				if condition.Status == isTrue && condition.Type == isComplete {
-					return true, nil
+				if (expected.Type != "" && condition.Type != expected.Type) ||
```
kononovn, line 369: Do we have any particular reason to implement such complicated if condition? Why do we need to change it ?

kononovn (changes requested):
Please fix few comments

pkg/cgu/cgu.go
klaskosk: @kononovn The diff is a bit weird here. I'm adding a new more general method `WaitForCondition` and then reimplementing the `WaitUntilComplete` function using `WaitForCondition` so the if statement handles more now. I'll use `WaitForCondition` directly in eco-gotests as well.

I've updated the PR to split the if statement into separate ones for clarity though.

kononovn (approved):
/lgtm

PR #483, opened 2024-06-17
argocd: reset application resource version on update

klaskosk:
When using force updates on an Argo CD application that has already been pulled from the cluster, there is a risk that the ResourceVersion is newer on the server but is also set in the Definition, resulting in the delete succeeding but not the creation during a force update.


kononovn (approved):
/lgtm

PR #497, opened 2024-07-04
configmap pkg: Update method added

elenagerman:
configmap pkg: Update method added


elenagerman:

elenagerman:

pkg/configmap/configmap_test.go
```diff
@@ -196,20 +195,34 @@ func TestDelete(t *testing.T) {
 	}
 }
 
-func TestWithOptions(t *testing.T) {
```
trewest, line 199: Seems you removed the withoptions test?

pkg/secret/secret.go
```diff
@@ -33,6 +33,12 @@ func NewBuilder(apiClient *clients.Settings, name, nsname string, secretType cor
 		"Initializing new secret structure with the following params: %s, %s, %s",
```
trewest, line 33: Please move secret updates to a separate PR

elenagerman: yea,it had to be a separate mr, it's by mistake here. thanks!

pkg/configmap/configmap_test.go
```diff
@@ -196,20 +196,34 @@ func TestDelete(t *testing.T) {
...
-		return builder, errors.New("error")
-	})
+	for _, testCase := range testCases {
+		testCase.testCM.WithData(testCase.testData)
+		_, err := testCase.testCM.Update()
 
-	assert.Equal(t, "error", testBuilder.errorMsg)
+		if testCase.expectedError != "" {
+			assert.Equal(t, testCase.expectedError, err.Error())
+		} else {
```
trewest, line 223: Can you change expectedError to an error type using fmt.Errorf and always check it (even if it is nil)?
```
```
Then just check the data if expectedError is nil
```
if testCase.expectedError == nil {
  assert.Equal(t, testCase.testData, testCase.testCM.Definition.Data)
}

Comment (elenagerman) [reply pkg/configmap/configmap_test.go:223]: @trewest I cannot use err nil verification in this case, cause this object is not really exists:
testCM:        buildTestBuilderWithFakeObjects([]runtime.Object{})

If you insist such verification has to be added I'll need to do a full unit-test refactoring

Comment (trewest) [reply pkg/configmap/configmap_test.go:223]: Yes, please make sure that the resource actually exists so we can test it thoroughly. Please look at the deployment test as an example - https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/deployment/deployment_test.go#L492

File: pkg/configmap/configmap_test.go [hunk 4]
```diff
@@ -196,20 +197,60 @@ func TestDelete(t *testing.T) {
...
+
+		// Set a value in the definition to test the update
+		testBuilder.Definition.Data = map[string]string{"key1": "value1", "key2": "value2"}
+
+		// Perform the update
+		result, err := testBuilder.Update()
+
+		// Assert the result
+		assert.NotNil(t, testBuilder.Definition)
+
```
trewest, line 245: Looks like you didn't use the expectedError parameter. Please just check assert.Equal(t, testCase.expectedError, err)

elenagerman: in provide by you example there is no such verification; I just remove expected error var at all

trewest (dismissed):
/lgtm

pkg/configmap/configmap.go
```diff
@@ -157,6 +157,32 @@ func (builder *Builder) Exists() bool {
 	return err == nil || !k8serrors.IsNotFound(err)
 }
 
+// Update renovates the existing kedaController object with kedaController definition in builder.
```
kononovn, line 160: Please fix the comment.

kononovn (changes requested):
1 small comment to fix.

kononovn (approved):
/lgtm

kononovn (approved):
/lgtm

PR #498, opened 2024-07-04
storage pkg: odf handler added, StorageSystem api object

elenagerman:
odf handler added, StorageSystem api object


elenagerman:

pkg/storage/odf.go
```diff
@@ -0,0 +1,297 @@
...
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	storageSystemObj := &odfoperatorv1alpha1.StorageSystem{}
+	err := builder.apiClient.Get(context.TODO(), goclient.ObjectKey{
+		Name:      builder.Definition.Name,
+		Namespace: builder.Definition.Namespace,
+	}, storageSystemObj)
+
+	if err != nil {
+		glog.V(100).Infof("storageSystem object %s does not exist in namespace %s",
```
trewest, line 131: This isn't necessarily an error due to the object not existing. I would just change the glog message to indicate that an error occurred

pkg/storage/odf.go
```diff
@@ -0,0 +1,297 @@
...
+// Delete removes storageSystem object from a cluster.
+func (builder *StorageSystemBuilder) Delete() error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof("Deleting the storageSystem object %s in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	err := builder.apiClient.Delete(context.TODO(), builder.Definition)
```
trewest, line 185: Check if the object exists first

pkg/storage/odf.go
elenagerman: changed, but I'm not so sure it is what you meant.

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+	for _, testCase := range testCases {
+		storageSystemObj, err := testCase.testStorageSystem.Get()
+
+		if testCase.expectedError == nil {
+			assert.Equal(t, storageSystemObj.Name, testCase.testStorageSystem.Definition.Name)
+			assert.Equal(t, storageSystemObj.Namespace, testCase.testStorageSystem.Definition.Namespace)
+			assert.Nil(t, err)
+		} else {
+			assert.Equal(t, testCase.expectedError.Error(), err.Error())
```
trewest, line 214: Can you move this above line 209 and always check it?

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+	for _, testCase := range testCases {
+		testStorageSystemBuilder, err := testCase.testStorageSystem.Create()
+
+		if testCase.expectedError == "" {
+			assert.Equal(t, testStorageSystemBuilder.Definition.Name, testStorageSystemBuilder.Object.Name)
+			assert.Equal(t, testStorageSystemBuilder.Definition.Namespace, testStorageSystemBuilder.Object.Namespace)
+			assert.Nil(t, err)
+		} else {
+			assert.Equal(t, testCase.expectedError, err.Error())
```
trewest, line 246: Please use error type for expectedError and check it unconditionally

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	for _, testCase := range testCases {
+		err := testCase.testStorageSystem.Delete()
+
+		if testCase.expectedError == nil {
+			assert.Nil(t, testCase.testStorageSystem.Object)
+			assert.Nil(t, err)
+		} else {
+			assert.Equal(t, testCase.expectedError.Error(), err.Error())
```
trewest, line 277: Consolidate the error checks above  line 273

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+		testCase.testStorageSystem.WithStorageClusterSpec(testCase.testKind, testCase.testName, testCase.testNamespace)
+		_, err := testCase.testStorageSystem.Update()
+
+		if testCase.expectedError == nil {
+			assert.Equal(t, testCase.testKind, testCase.testStorageSystem.Definition.Spec.Kind)
+			assert.Equal(t, testCase.testName, testCase.testStorageSystem.Definition.Spec.Name)
+			assert.Equal(t, testCase.testNamespace, testCase.testStorageSystem.Definition.Spec.Namespace)
+		} else {
+			assert.Equal(t, testCase.expectedError, err)
```
trewest, line 320: Same

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,401 @@
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		testBuilder := buildValidStorageSystemBuilder(buildStorageSystemClientWithDummyObject())
+
+		result := testBuilder.WithStorageClusterSpec(testCase.testKind, testCase.testName, testCase.testNamespace)
+
+		if testCase.expectedError != "" {
+			assert.Equal(t, testCase.expectedError, result.errorMsg)
```
trewest, line 364: Check it unconditionally

pkg/storage/odf_test.go
elenagerman: @trewest please, let's leave it like this. It is imposible to use it generic. I need error message, I cannot compare error with error they are not the same. I need to verify error msg is as expected. But,when error is not expected == nil there is no object to get msg from and there is panic failure:
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x212c940]

pkg/storage/odf_test.go
elenagerman: imposible

pkg/storage/odf_test.go
elenagerman: same

pkg/storage/odf_test.go
elenagerman: same

pkg/storage/odf_test.go
trewest: Panics should only occur if you call the Error() method on an error that is nil. It isn't necessary for determining error messages, you just have to compare the error objects directly `assert.Equal(t,  testCase.expectedError, err)`. Please let me know if that example doesn't work in your scenario

elenagerman: in this specific case it is work, but I changed an expectedError type

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,398 @@
...
+			testSettings = clients.GetTestClients(clients.TestClientParams{
+				K8sMockObjects: runtimeObjects,
+			})
+		}
+
+		builderResult, err := PullStorageSystem(testSettings, testCase.name, testCase.namespace)
+		assert.Equal(t, testCase.expectedError, err)
+
+		if testCase.expectedError != nil {
+			assert.Equal(t, testCase.expectedError.Error(), err.Error())
```
trewest, line 96: Can be removed

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,398 @@
...
+		} else {
+			assert.Equal(t, testStorageSystem.Name, builderResult.Object.Name)
+			assert.Equal(t, testStorageSystem.Namespace, builderResult.Object.Namespace)
+			assert.Nil(t, err)
```
trewest, line 100: Can be removed

pkg/storage/odf_test.go
trewest: Why is this impossible? Isn't it the same as the other cases I mentioned?

pkg/storage/odf_test.go
trewest: Again, I think it should be the same as the other cases

pkg/storage/odf_test.go
elenagerman: I cannot parse error message in this case, only if I compare the whole error.
In this specific case it is working, but there are many it will fail

pkg/storage/odf_test.go
elenagerman: this is for an update:
`--- PASS: TestStorageSystemDelete (0.01s)
=== RUN   TestStorageSystemUpdate
    odf_test.go:311: 
                Error Trace:    /home/elgerman/eco-gotest/eco-goinfra-odf/pkg/storage/odf_test.go:311
                Error:          Not equal: 
                                expected: <nil>(<nil>)
                                actual  : *errors.StatusError(&errors.StatusError{ErrStatus:v1.Status{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ListMeta:v1.ListMeta{SelfLink:"", ResourceVersion:"", Continue:"", RemainingItemCount:(*int64)(nil)}, Status:"Failure", Message:"Operation cannot be fulfilled on storagesystems.odf.openshift.io \"ocs-storagecluster-storagesystem\": object was modified", Reason:"Conflict", Details:(*v1.StatusDetails)(0xc00094bb00), Code:409}})
                Test:           TestStorageSystemUpdate
--- FAIL: TestStorageSystemUpdate (0.01s)
`

trewest: Yeah it seems to be because your Update method doesn't have the force argument. Similar to other resources that use the runtime-client. E.g. https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/cgu/cgu.go#L250

trewest: This means that update probably won't work in a real environment also

elenagerman: but it is working :) this is not the first time I see such error.
but let me see what I can do...

elenagerman: @trewest if what you are saying is right, so:
- there is no other option then to add force as hardcoded, not force doesn't work at all; please confirm
- you need to fix TestClusterImageSetUpdate test, since there is no force=true variation included

elenagerman: @trewest another option is just to delete this method

trewest (approved):
/lgtm

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// StorageSystemBuilder provides struct for storageSystem object containing connection
+// to the cluster and the storageSystem definitions.
+type StorageSystemBuilder struct {
```
kononovn, line 20: package is name storage. Let remove storage from the object name in order to remove duplication.
I think we also need to add odf in order to point to specific storage class.
Please re-name it to:
OdfSystemBuilder

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+	Object *odfoperatorv1alpha1.StorageSystem
+	// api client to interact with the cluster.
+	apiClient goclient.Client
+	// Used in functions that define or mutate storageSystem definition. errorMsg is processed before the
+	// storageSystem object is created.
+	errorMsg string
+}
+
+// NewStorageSystemBuilder creates a new instance of Builder.
+func NewStorageSystemBuilder(apiClient *clients.Settings, name, nsname string) *StorageSystemBuilder {
```
kononovn, line 33: Same here. Please re-name to NewOdfSystemBuilder

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+		builder.errorMsg = "storageSystem 'nsname' cannot be empty"
+
+		return builder
+	}
+
+	return builder
+}
+
+// PullStorageSystem gets an existing storageSystem object from the cluster.
+func PullStorageSystem(apiClient *clients.Settings, name, namespace string) (*StorageSystemBuilder, error) {
```
kononovn, line 73: PullOdfSystem

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,351 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var (
+	defaultStorageSystemName      = "ocs-storagecluster-storagesystem"
+	defaultStorageSystemNamespace = "openshift-storage"
+)
+
+func TestSorageSystemPull(t *testing.T) {
```
kononovn, line 20: Don't forget to re-name test functions as well.

pkg/storage/odf_test.go
```diff
@@ -0,0 +1,351 @@
...
+			} else {
+				assert.Nil(t, testStorageSystemBuilder)
+			}
+		} else {
+			assert.Equal(t, testCase.expectedError, testStorageSystemBuilder.errorMsg)
+			assert.NotNil(t, testStorageSystemBuilder.Definition)
+		}
+	}
+}
+
```
kononovn, line 157: I noticed that Update function is missing. Can you add it?

kononovn (changes requested):
Looks good, few comments to fix.

pkg/storage/odf.go
elenagerman: and both of the odf related

pkg/storage/odf_test.go
elenagerman: it is not relevant in this case. You cannot update. you can only delete and create new

pkg/storage/odf.go
kononovn: Please refer to https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/storage/storageclass.go . The same logic we removed storage from object name in order to avoid duplication. Can we do the same wit this odf object?

kononovn (changes requested):
looks good. Just 1 last detail to fix: https://github.com/openshift-kni/eco-goinfra/pull/498#discussion_r1679610648

pkg/storage/odf.go
elenagerman: same here

pkg/storage/odf.go
elenagerman: same here

elenagerman:
@kononovn there is no any logic to remove "storage" from the methods name. It will be hard to find the correct method after.

to use " PullODFSystem" when you need to pull StorageSystem obj???

+ we have ocs file which is handling StorageClusterBuilder, so what is a problem with this one???
storage package is growing and it will be very hard to find there anything especially when it does not have a correct names

pkg/storage/odf.go
elenagerman: this is a point, I tried to explain to you when I worked on storageclass handler as well.
never mind, as you wish ...

kononovn:
> @kononovn there is no any logic to remove "storage" from the methods name. It will be hard to find the correct method after.
> 
> to use " PullODFSystem" when you need to pull StorageSystem obj???
> 
> * we have ocs file which is handling StorageClusterBuilder, so what is a problem with this one???
>   storage package is growing and it will be very hard to find there anything especially when it does not have a correct names

It'll be easy to understand that this resource is belong to ODF as well as to storage. Please check few examples below:
Here is my proposal:
storageSystem := storage.NewSystemODF()
storageSystem :=storage.PullSystemODF()
As you can see It's easy to understand that package belongs to storage package and it's part of ODF operator

Current state from :
storageSystem := storage.NewSystemBuilder()
storageSystem := storage.PullSystem()
It's hard to understand that resource belongs to ODF operator.

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+	apiClient goclient.Client
+	// Used in functions that define or mutate storageSystem definition. errorMsg is processed before the
+	// storageSystem object is created.
+	errorMsg string
+}
+
+// NewSystemODFBuilder creates a new instance of Builder.
+func NewSystemODFBuilder(apiClient *clients.Settings, name, nsname string) *SystemODFBuilder {
+	glog.V(100).Infof(
+		"Initializing new storageSystem structure with the following params: %s, %s", name, nsname)
```
kononovn, line 35: new odf storageSystem

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if apiClient == nil {
+		glog.V(100).Infof("storageSystem 'apiClient' cannot be empty")
```
kononovn, line 38: Can you please fix storageSystem to SystemODF in all comments/logs below

kononovn (changes requested):
Please fix few cosmetic issues.

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+
+// WithStorageClusterSpec sets the storage SystemODF with storageCluster spec values.
+func (builder *SystemODFBuilder) WithStorageClusterSpec(
+	kind odfoperatorv1alpha1.StorageKind, name, nsname string) *SystemODFBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
+	glog.V(100).Infof(
+		"Setting storage SystemODF %s in namespace %s with storageCluster spec; \n"+
```
kononovn, line 214: SystemODF

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+			"kind: %v, name: %s, namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace, kind, name, nsname)
+
+	if kind == "" {
+		glog.V(100).Infof("The kind of the storageCluster spec is empty")
```
kononovn, line 219: Please change to SystemODF

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+
+		builder.errorMsg = "storageCluster spec 'kind' cannot be empty"
```
kononovn, line 221: Please change to SystemODF

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+
+		return builder
+	}
+
+	if name == "" {
+		glog.V(100).Infof("The name of the storageCluster spec is empty")
```
kononovn, line 227: Please change to SystemODF

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+
+		builder.errorMsg = "storageCluster spec 'name' cannot be empty"
```
kononovn, line 229: Please change to SystemODF  all logs/comments below

pkg/storage/odf.go
```diff
@@ -0,0 +1,279 @@
...
+	builder.Definition.Spec.Name = name
+	builder.Definition.Spec.Namespace = nsname
+
+	return builder
+}
+
+// validate will check that the builder and builder definition are properly initialized before
+// accessing any member fields.
+func (builder *SystemODFBuilder) validate() (bool, error) {
+	resourceCRD := "StorageSystem"
```
kononovn, line 252: Please change to SystemODF

kononovn (changes requested):
Few cosmetic issues.

pkg/storage/odf.go
elenagerman: the method name is "WithStorageClusterSpec", it includes storagesystemCluster definition, not storage system definition

pkg/storage/odf.go
elenagerman: same

pkg/storage/odf.go
elenagerman: this is not storageODF name, this is storageCluster name

kononovn (approved):
/lgtm

PR #499, opened 2024-07-05
localvolumediscovery enhancements, unit test, WithNodeSelector added

elenagerman:
lso pkg: localvolumediscovery enhancements + unit test + WithNodeSelector method added

### DIFF
--- a/pkg/lso/localvolumediscovery.go
+++ b/pkg/lso/localvolumediscovery.go
@@ -3,13 +3,18 @@ package lso
 import (
 	"context"
 	"fmt"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	corev1 "k8s.io/api/core/v1"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/golang/glog"
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
 	"github.com/openshift-kni/eco-goinfra/pkg/msg"
-	lsoV1alpha1 "github.com/openshift/local-storage-operator/api/v1alpha1"
+	lsov1alpha1 "github.com/openshift/local-storage-operator/api/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	goclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -18,14 +23,14 @@ import (
 // and a localVolumeDiscovery definition.
 type LocalVolumeDiscoveryBuilder struct {
 	// localVolumeDiscovery definition, used to create the localVolumeDiscovery object.
-	Definition *lsoV1alpha1.LocalVolumeDiscovery
+	Definition *lsov1alpha1.LocalVolumeDiscovery
 	// Created localVolumeDiscovery object.
-	Object *lsoV1alpha1.LocalVolumeDiscovery
+	Object *lsov1alpha1.LocalVolumeDiscovery
 	// Used in functions that define or mutate localVolumeDiscovery definition. errorMsg is processed
 	// before the localVolumeDiscovery object is created
 	errorMsg string
 	// api client to interact with the cluster.
-	apiClient *clients.Settings
+	apiClient goclient.Client
 }
 
 // NewLocalVolumeDiscoveryBuilder creates new instance of LocalVolumeDiscoveryBuilder.
@@ -33,9 +38,15 @@ func NewLocalVolumeDiscoveryBuilder(apiClient *clients.Settings, name, nsname st
 	glog.V(100).Infof("Initializing new localVolumeDiscovery structure with the following params: name: "+
 		"%s, namespace: %s", name, nsname)
 
-	builder := LocalVolumeDiscoveryBuilder{
-		apiClient: apiClient,
-		Definition: &lsoV1alpha1.LocalVolumeDiscovery{
+	if apiClient == nil {
+		glog.V(100).Infof("localVolumeDiscovery 'apiClient' cannot be empty")
+
+		return nil
+	}
+
+	builder := &LocalVolumeDiscoveryBuilder{
+		apiClient: apiClient.Client,
+		Definition: &lsov1alpha1.LocalVolumeDiscovery{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      name,
 				Namespace: nsname,
@@ -47,15 +58,19 @@ func NewLocalVolumeDiscoveryBuilder(apiClient *clients.Settings, name, nsname st
 		glog.V(100).Infof("The name of the localVolumeDiscovery is empty")
 
 		builder.errorMsg = "localVolumeDiscovery 'name' cannot be empty"
+
+		return builder
 	}
 
 	if nsname == "" {
 		glog.V(100).Infof("The nsname of the localVolumeDiscovery is empty")
 
 		builder.errorMsg = "localVolumeDiscovery 'nsname' cannot be empty"
+
+		return builder
 	}
 
-	return &builder
+	return builder
 }
 
 // PullLocalVolumeDiscovery retrieves an existing localVolumeDiscovery object from the cluster.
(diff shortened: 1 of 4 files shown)


elenagerman:

pkg/lso/localvolumediscovery.go
```diff
@@ -223,11 +212,49 @@ func (builder *LocalVolumeDiscoveryBuilder) Update() (*LocalVolumeDiscoveryBuild
...
+		return false, err
+	}
+
+	glog.V(100).Infof("Verify localVolumeDiscovery %s in namespace %s is in Discovering phase",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if !builder.Exists() {
+		return false, fmt.Errorf("localVolumeDiscovery %s not found in %s namespace",
+			builder.Definition.Name, builder.Definition.Namespace)
+	}
```
trewest, line 239: I think you can omit this since GetPhase already checks if the resource exists

pkg/lso/localvolumediscovery_test.go
```diff
@@ -0,0 +1,404 @@
...
+			testSettings = clients.GetTestClients(clients.TestClientParams{
+				K8sMockObjects: runtimeObjects,
+			})
+		}
+
+		builderResult, err := PullLocalVolumeDiscovery(testSettings, testCase.name, testCase.namespace)
+		assert.Equal(t, testCase.expectedError, err)
+
+		if testCase.expectedError != nil {
+			assert.Equal(t, testCase.expectedError.Error(), err.Error())
```
trewest, line 97: Check errors unconditionally

pkg/lso/localvolumediscovery_test.go
```diff
@@ -0,0 +1,404 @@
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		localVolumeDiscoveryObj, err := testCase.testLocalVolumeDiscovery.Get()
+
+		if testCase.expectedError == nil {
+			assert.Equal(t, localVolumeDiscoveryObj.Name, testCase.testLocalVolumeDiscovery.Definition.Name)
+			assert.Equal(t, localVolumeDiscoveryObj.Namespace, testCase.testLocalVolumeDiscovery.Definition.Namespace)
+			assert.Nil(t, err)
```
trewest, line 212: Check error unconditionally

pkg/lso/localvolumediscovery_test.go
```diff
@@ -0,0 +1,404 @@
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		testLocalVolumeSetBuilder, err := testCase.testLocalVolumeDiscovery.Create()
+
+		if testCase.expectedError == "" {
+			assert.Equal(t, testLocalVolumeSetBuilder.Definition.Name, testLocalVolumeSetBuilder.Object.Name)
+			assert.Equal(t, testLocalVolumeSetBuilder.Definition.Namespace, testLocalVolumeSetBuilder.Object.Namespace)
+			assert.Nil(t, err)
```
trewest, line 244: Check error unconditionally

pkg/lso/localvolumediscovery_test.go
```diff
@@ -0,0 +1,404 @@
...
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		assert.Nil(t, nil, testCase.testLocalVolumeDiscovery.Object)
+		testCase.testLocalVolumeDiscovery.WithNodeSelector(testCase.testNodeSelector)
+		_, err := testCase.testLocalVolumeDiscovery.Update()
+
+		if testCase.expectedError != "" {
```
trewest, line 304: Check error unconditionally

pkg/lso/localvolumediscovery_test.go
```diff
@@ -0,0 +1,404 @@
...
+			expectedErrorText: "",
+		},
+	}
+
+	for _, testCase := range testCases {
+		testBuilder := buildValidLVDObjectBuilder(buildLVDClientWithDummyObject())
+
+		result := testBuilder.WithNodeSelector(testCase.testNodeSelector)
+
+		if testCase.expectedError {
```
trewest, line 366: Check error unconditionally, you can just use one variable for the expected error

pkg/lso/localvolumediscovery_test.go
elenagerman: @trewest I cannot check it unconditionally in this case, I need an access to the error message to compare and when there is no error I'm failing on panic: runtime error: invalid memory address or nil pointer dereference

pkg/lso/localvolumediscovery_test.go
elenagerman: @trewest it isimpossible here as weel. The reason is different. For the Update method we always have a "conflict' error on demo client:
Message:"Operation cannot be fulfilled on localvolumediscoveries.local.storage.openshift.io \"auto-discover-devices\": object was modified", Reason:"Conflict"

pkg/lso/localvolumediscovery_test.go
elenagerman: @trewest sameproblem on getting error msg

pkg/lso/localvolumediscovery_test.go
trewest: This assertion doesn't work?
```
```

elenagerman:

pkg/lso/localvolumediscovery.go
```diff
@@ -184,50 +181,62 @@ func (builder *LocalVolumeDiscoveryBuilder) Delete() error {
...
 
-	return builder, err
+				return false, nil
+			}
+
+			if phase == "Discovering" {
+				return true, nil
+			}
+
+			return false, nil
```
trewest, line 225:
```suggestion
			return phase == "Discovering", nil
```

trewest (approved):
/lgtm

pkg/lso/localvolumediscovery_test.go
```diff
@@ -0,0 +1,391 @@
...
+		testLocalVolumeDiscovery *LocalVolumeDiscoveryBuilder
+		expectedError            error
+	}{
+		{
+			testLocalVolumeDiscovery: buildValidLVDObjectBuilder(buildLVDClientWithDummyObject()),
+			expectedError:            nil,
+		},
+		{
+			testLocalVolumeDiscovery: buildValidLVDObjectBuilder(clients.GetTestClients(clients.TestClientParams{})),
+			expectedError:            fmt.Errorf("localVolumeDiscovery cannot be deleted because it does not exist"),
```
kononovn, line 257: Please fix Delete function. We don't return error if Object doesn't exist during Delete call

pkg/lso/localvolumediscovery_test.go
```diff
@@ -0,0 +1,391 @@
...
+	for _, testCase := range testCases {
+		err := testCase.testLocalVolumeDiscovery.Delete()
+		assert.Equal(t, testCase.expectedError, err)
+
+		if testCase.expectedError == nil {
+			assert.Nil(t, testCase.testLocalVolumeDiscovery.Object)
+		}
+	}
+}
+
```
kononovn, line 270: Please add test case for IsDiscovering function

kononovn (changes requested):
2 comments only.

pkg/lso/localvolumediscovery_test.go
elenagerman: it's a dummy object, how can I really test discovering state on not real object, it's a status, not spec.
+ it should be other object types involed to test it is really discovering

elenagerman: such test is possible only in case I will replace "return builder.Object.Status.Phase, nil" to the "return builder.Definition.Status.Phase, nil" (Object to the Definition). But I am not sure, that in this case this method will work with the real object

pkg/lso/localvolumediscovery_test.go
kononovn: Please fix Delete function

kononovn (changes requested):
Please fix last comment related to Delete function

pkg/lso/localvolumediscovery_test.go
elenagerman: yes, sorry, my laptop crashed when I worked on this change. I was sure it is in already

pkg/lso/localvolumediscovery.go
```diff
@@ -170,64 +167,72 @@ func (builder *LocalVolumeDiscoveryBuilder) Delete() error {
...
 	if !builder.Exists() {
-		return fmt.Errorf("localVolumeDiscovery cannot be deleted because it does not exist")
+		return nil
 	}
 
 	err := builder.apiClient.Delete(context.TODO(), builder.Definition)
 
 	if err != nil {
-		return fmt.Errorf("can not delete localVolumeDiscovery: %w", err)
+		return nil
```
kononovn, line 176: Please revert this back. We actually want to catch  real error if occurred

pkg/lso/localvolumediscovery.go
```diff
@@ -170,64 +167,72 @@ func (builder *LocalVolumeDiscoveryBuilder) Delete() error {
 		builder.Definition.Name, builder.Definition.Namespace)
 
 	if !builder.Exists() {
-		return fmt.Errorf("localVolumeDiscovery cannot be deleted because it does not exist")
+		return nil
```
kononovn, line 170: Please add log message above:
glog.V(100).Infof("localVolumeDiscovery %s in namespace %s cannot be deleted because it does not exist",

kononovn (changes requested):

kononovn (approved):
/lgtm

PR #507, opened 2024-07-12
BMC: added redfish api to get and change system's boot order.

greyerof:
- SystemBootOptions: helper function that reads the current available boot options and creates a map so the user can see the display name of each of boot option reference.
- SystemBootOrderReferences: returns a slice with the current boot order references.
- SetSystemBootOrderReferences: updates the system's boot order references. Needs a system reset so changes take effect.


klaskosk (approved):

kononovn (approved):
/lgtm

PR #511, opened 2024-07-14
storage pkg: PVC handler adjustments + unit test

elenagerman:


elenagerman:

pkg/storage/pvc_test.go
```diff
@@ -0,0 +1,441 @@
...
+		testPVC       *PVCBuilder
+		expectedError error
+	}{
+		{
+			testPVC:       buildValidPVCTestBuilder(buildTestClientWithDummyPVC()),
+			expectedError: nil,
+		},
+		{
+			testPVC:       buildInvalidPVCTestBuilder(buildTestClientWithDummyPVC()),
+			expectedError: fmt.Errorf("invalid builder"),
```
kononovn, line 199: Please fix Create method.
It has to **return builder,err** 
instead of
**return builder, fmt.Errorf("invalid builder")**

kononovn (changes requested):
1 comment to fix.

pkg/storage/pvc_test.go
elenagerman: sorry, I do not understand your point

kononovn: Current state of Create function

// Create generates a PVC in cluster and stores the created object in struct.
func (builder *PVCBuilder) Create() (*PVCBuilder, error) {
		return builder, **fmt.Errorf("invalid builder")**
	}

Required state:

```
// Create generates a PVC in cluster and stores the created object in struct.
func (builder *PVCBuilder) Create() (*PVCBuilder, error) {
		return builder, err
	}
```

kononovn (changes requested):
Please check comment above

kononovn (approved):
/lgtm

PR #523, opened 2024-07-18
sync: add excludes field for repo

trewest:
Adds excludes fields for sync configs. Allows us to remove files and directories from projects we clone into our repository. Useful for cases where we want to remove unneeded files or remove go.mod and vendored files in favor of our own cloned files.


achuzhoy (approved):

kononovn (approved):
/lgtm

PR #526, opened 2024-07-22
Add modifiable scheme to client settings struct

trewest:
- Modifies Settings struct to contain trackable/modifiable scheme field
- Enables us to move scheme initialization from clients pkg to each resources' respective pkg
- Allows the client to be more lightweight and only aware of resource schemes upon accessing a resources' NewBuilder or Pull functions

### DIFF
--- a/pkg/clients/clients.go
+++ b/pkg/clients/clients.go
@@ -160,7 +160,11 @@ type Settings struct {
 	clientCguV1.RanV1alpha1Interface
 	ClusterClient clusterClient.Interface
 	clusterV1Client.ClusterV1Interface
+	scheme *runtime.Scheme
 }
+
+// SchemeAttacher represents a function that can modify the clients current schemes.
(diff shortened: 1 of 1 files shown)


pkg/clients/clients.go
```diff
@@ -444,10 +444,30 @@ func (settings *Settings) GetAPIClient() (*Settings, error) {
...
+	err := attacher(settings.scheme)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// SchemeAttacher represents a function that can modify the clients current schemes.
+type SchemeAttacher func(*runtime.Scheme) error
```
kononovn, line 464: Can you please move this public struct after Settings?

kononovn:
@trewest Can you change PR header/commit message to:
client: Add modifiable scheme to settings struct

kononovn (approved):
/lgtm

PR #528, opened 2024-07-22
metallb: switched to runtime client

kononovn:


pkg/metallb/l2advertisement.go
```diff
@@ -44,12 +39,8 @@ func NewL2AdvertisementBuilder(apiClient *clients.Settings, name, nsname string)
 	}
```
trewest, line 40: Do you need to add this to the apiClient scheme?

pkg/metallb/l2advertisement.go
```diff
@@ -83,7 +74,7 @@ func PullL2Advertisement(apiClient *clients.Settings, name, nsname string) (*L2A
...
```
trewest, line 75: Do you need to add this to the apiClient scheme?

pkg/metallb/addresspool.go
```diff
@@ -38,13 +33,14 @@ func NewIPAddressPoolBuilder(
 		"Initializing new IPAddressPool structure with the following params: %s, %s %s",
 		name, nsname, addrPool)
 
+	err := apiClient.AttachScheme(mlbtypes.AddToScheme)
+	if err != nil {
+		return nil
```
trewest, line 38: I would add a glog in case this happens. It may have confusing behavior otherwise

trewest (approved):
/lgtm

PR #530, opened 2024-07-23
sync: refactor sync process

klaskosk:
Previously, the sync process would attempt to reverse the changes to be applied to the local directory and diff it against the cloned directory. If they were the same, the local directory would be reverted using git. If they were different, the cloned directory would have the opposite changes applied and then be copied to the local directory.

This relied on the local directory already being in git and would cause errors when it was not tracked in git. This PR changes this process to apply the desired updates to the cloned directory, diff it against the local, and overwrite the local if there is a difference. This does not rely on git to revert the changes and should halve the number of changes to files.

### DIFF
--- a/internal/sync/sync.go
+++ b/internal/sync/sync.go
@@ -58,14 +58,16 @@ func syncRemoteRepo(repo *repo) {
 	projectLocalDirectory := path.Join("./pkg", repo.LocalAPIDirectory)
 
 	gitClone(basePath, repo)
+	excludeAndRefactor(projectClonedDirectory, projectLocalDirectory, repo)
 
 	glog.V(100).Infof("Comparing local %s and cloned %s api directories for repo %s",
 		projectLocalDirectory, projectClonedDirectory, repo.Name)
 
-	if repoSynced(projectClonedDirectory, projectLocalDirectory, repo) {
-		gitReset(repo.LocalAPIDirectory)
-	} else {
-		syncDirectories(projectClonedDirectory, projectLocalDirectory, repo)
+	err := execCmd("", "diff", []string{projectClonedDirectory, projectLocalDirectory})
+	if err != nil {
+		glog.V(100).Infof("Repos not synced. Copying cloned repo %s to %s", projectClonedDirectory, projectLocalDirectory)
+
+		copyClonedToLocal(projectClonedDirectory, projectLocalDirectory)
 	}
 
 	glog.V(100).Infof("Remove cloned directory from filesystem: %s", path.Join(basePath, repo.Name))
@@ -81,21 +83,13 @@ func syncRemoteRepo(repo *repo) {
 	}
 }
 
-func repoSynced(clonedDir, localDir string, repo *repo) bool {
-	glog.V(100).Infof("Verifying destination directory %s exists", localDir)
-
-	if _, err := os.Stat(localDir); os.IsNotExist(err) {
-		glog.V(100).Infof("Destination api directory %s doesn't exist creating directory", localDir)
-
-		if os.MkdirAll(localDir, 0777) != nil {
-			glog.V(100).Infof("Failed to create api directory. Exit with error code 1")
-			os.Exit(1)
-		}
-	}
+// excludeAndRefactor excludes and refactors files in the clonedDir to prepare them for being compared or copied to the
+// localDir.
+func excludeAndRefactor(clonedDir, localDir string, repo *repo) {
+	glog.V(100).Infof("Updating %s to match expected state of %s", clonedDir, localDir)
 
 	if len(repo.Excludes) > 0 {
-		glog.V(100).Infof("Remove excluded files under %s",
-			path.Base(clonedDir))
+		glog.V(100).Infof("Remove excluded files under %s", path.Base(clonedDir))
 
 		err := excludeFiles(clonedDir, repo.Excludes...)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


internal/sync/sync.go
```diff
@@ -383,36 +397,28 @@ func refactorFunc(oldLine, newLine string, filePatterns []string) filepath.WalkF
...
-	entries, err := os.ReadDir(path)
-	if err != nil {
-		return err
-	}
+	return filepath.WalkDir(path, excludeFunc(path, patterns))
+}
 
-	for _, entry := range entries {
-		joinedPath := filepath.Join(path, entry.Name())
+func excludeFunc(root string, patterns []string) fs.WalkDirFunc {
```
trewest, line 403: This works when removing directories?

klaskosk: @trewest I've tried it and it seems to work fine. [Looking at the WalkDir source](https://cs.opensource.google/go/go/+/refs/tags/go1.22.5:src/path/filepath/path.go;l=441-473) it always calls the function on a path before reading it so this does result in `excludeFunc` being run on a path that does not exist, but since `os.RemoveAll` returns nil on paths that don't exist it'll be fine.

kononovn:
@klaskosk 

`When syncing a new repo using the internal/sync package, subsequent syncs will fail when the original sync is not tracked by git and no changes are necessary. This causes issues when working on adding a new repo and only one local directory needs to be updated. Rather than failing the entire sync as it currently does, this PR allows it to recover and move on to other repos.`

This feature is not needed. Original sync has to be added manually by package owner. It's OK to fail sync even if single package failed. It means we need to debug it immediately instead of ignoring the issue.

kononovn:
> @klaskosk
> 
> `When syncing a new repo using the internal/sync package, subsequent syncs will fail when the original sync is not tracked by git and no changes are necessary. This causes issues when working on adding a new repo and only one local directory needs to be updated. Rather than failing the entire sync as it currently does, this PR allows it to recover and move on to other repos.`
> 
> This feature is not needed. Original sync has to be added manually by package owner. It's OK to fail sync even if single package failed. It means we need to debug it immediately instead of ignoring the issue.

Ok may be I missed the point. Let me checkout locally and review

internal/sync/sync.go
trewest: Okay, it kept failing for me for some reason when I tried to implement it this way. Thanks

trewest:
> @klaskosk
> 
> `When syncing a new repo using the internal/sync package, subsequent syncs will fail when the original sync is not tracked by git and no changes are necessary. This causes issues when working on adding a new repo and only one local directory needs to be updated. Rather than failing the entire sync as it currently does, this PR allows it to recover and move on to other repos.`
> 
> This feature is not needed. Original sync has to be added manually by package owner. It's OK to fail sync even if single package failed. It means we need to debug it immediately instead of ignoring the issue.

@kononovn Is there any harm in making the script handle scenarios where we want to sync multiple times without failing or committing? IMO it makes developing an adequate config a bit cumbersome. The user has to continuously remove the cloned directory or commit changes rather than just re-running `make lib-sync`.

kononovn:
> g a bit cumbersome. The user has to continuously remove the cloned directory or commit changes rather

Agree, as I mentioned before I missed few points. Testing locally

trewest:
@klaskosk @kononovn To save time, can we split it into two PRs? The excludes updates look good to me and I could use them for something I'm currently working on.

klaskosk:
@trewest sure I can make those a separate PR

klaskosk:
Okay so I've realized that there is a flaw in this PR since the process of checking if the repos are synced involves changing the local dir so although this removes the error it does not allow you to properly sync over old changes

klaskosk:
@kononovn @trewest I've gone ahead and updated the PR and its description. It now refactors the sync process to not rely on git for resetting files rather than making an exception for untracked files.

After the changes, the output on the current configs is identical to the old process so it doesn't seem to break anything existing in the process.

internal/sync/sync.go
```diff
@@ -108,115 +100,44 @@ func repoSynced(clonedDir, localDir string, repo *repo) bool {
...
-
-	glog.V(100).Infof("Copy api filed from cloned directory to local api directory")
-
-	err := execCmd(
-		"",
-		"cp",
-		[]string{"-a", fmt.Sprintf("%s/.", clonedDir), fmt.Sprintf("%s/", localDir)})
-
+	// We do MkdirAll even though we do RemoveAll so that the path to localDir exists.
+	err := os.RemoveAll(localDir)
```
trewest, line 130: Can you explain why we create the directory and then remove the directory?

trewest: Not sure I follow based on the comment alone

klaskosk: @trewest We create it to make sure the path leading up to it exists and then just delete the specific localDir only. I updated the comments to make this more clear

trewest (approved):
/lgtm

internal/sync/sync.go
```diff
@@ -81,21 +83,11 @@ func syncRemoteRepo(repo *repo) {
...
-
-	if _, err := os.Stat(localDir); os.IsNotExist(err) {
-		glog.V(100).Infof("Destination api directory %s doesn't exist creating directory", localDir)
-
-		if os.MkdirAll(localDir, 0777) != nil {
-			glog.V(100).Infof("Failed to create api directory. Exit with error code 1")
-			os.Exit(1)
-		}
-	}
+func excludeAndRefactor(clonedDir, localDir string, repo *repo) {
```
kononovn, line 86: Please either add comment to this function:
// Remove excludes files and refactor code in cloned directory. 

or re-name it to excludeFilesAndRefactorCodeInCloneDirectory

kononovn (changes requested):
1 minor comment. It'll improve readability

kononovn (approved):
/lgtm

PR #532, opened 2024-07-24
sync: fix excludes for nested dirs

klaskosk:
The recursion in excludeFiles abandons searching at the current level in favor of searching a subdirectory when one is found. For example, if there is a directory `dir`, with subdirectories `a` and `b` which have `a_test.go` and `b_test.go`, respectively, the current implementation will walk `dir`, then `a`, remove `a_test.go`, and return. This would skip checking `b`, however, and has been changed to use filepath.WalkDir and just ignore the root directory to avoid that scenario.

Additionally, refactor has been updated to use filepath.WalkDir since it did not need the extra information from filepath.Walk that makes filepath.Walk less efficient.

### DIFF
--- a/internal/sync/sync.go
+++ b/internal/sync/sync.go
@@ -3,6 +3,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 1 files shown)


klaskosk:
@trewest Here's the PR with just the excludes fix

trewest (approved):
/lgtm

kononovn (approved):
/lgtm

PR #534, opened 2024-07-24
nvidiagpu: switch to internal packages

kononovn:


pkg/nvidiagpu/clusterpolicy.go
```diff
@@ -4,10 +4,10 @@ import (
 	"context"
 	"fmt"
 
-	nvidiagpuv1 "github.com/NVIDIA/gpu-operator/api/v1"
 	"github.com/golang/glog"
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
 	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	nvidiagpuv1 "github.com/openshift-kni/eco-goinfra/pkg/schemes/nvidiagpu/nvidiagputypes"
```
trewest, line 10: Do you want to also move AddToScheme here?

kononovn: I'll add it to the next PR

PR #536, opened 2024-07-24
argocd: move package to internal schemes

klaskosk:
Switches to use internal schemes for both of the argocd builders, including switching the ApplicationBuilder to the runtime client.


klaskosk:
I've updated this PR but it is currently awaiting #590 before it can be merged

kononovn (dismissed):
/lgtm

pkg/argocd/applications.go
```diff
@@ -44,8 +35,15 @@ func PullApplication(apiClient *clients.Settings, name, nsname string) (*Applica
 		return nil, fmt.Errorf("application 'apiClient' cannot be empty")
 	}
 
+	err := apiClient.AttachScheme(argocdtypes.AddToScheme)
+	if err != nil {
+		glog.V(100).Info("Failed to add Application scheme to client schemes")
```
trewest, line 40: Maybe include argocd in the message so it is clear which application resource the used is being notified of
```suggestion
		glog.V(100).Info("Failed to add argocd application scheme to client schemes")
```

pkg/argocd/applications.go
```diff
@@ -198,34 +195,18 @@ func (builder *ApplicationBuilder) Create() (*ApplicationBuilder, error) {
...
-		}
+	err := builder.apiClient.Create(context.TODO(), builder.Definition)
+	if err != nil {
+		return nil, err
 	}
 
-	return builder, err
+	builder.Object = builder.Definition
+
+	return builder, nil
```
trewest, line 209: Instead of having multiple returns, you can do something like this:
```
	if !builder.Exists() {
		err = builder.apiClient.Create(context.TODO(), builder.Definition)
		if err == nil {
		}
	}

	return builder, err
```

klaskosk:
@trewest I've updated the PR to address the comments

trewest (approved):
/lgtm

PR #546, opened 2024-07-30
oadp: add sync config and internal schemes

trewest:
Adds a sync config for OADP along with the relevant internal pkgs. 

OADP automatic sync is disabled due to complexities with cloning the project into eco-goinfra. Without manual intervention, there are too many dependencies and unneeded changes.


kononovn (approved):
/lgtm

PR #553, opened 2024-07-30
ocm: added unit tests and increased policy coverage

klaskosk:
This PR adds unit tests for list and validate for the policy resource in the OCM package.

### DIFF
--- a/pkg/ocm/policy.go
+++ b/pkg/ocm/policy.go
@@ -23,7 +23,7 @@ type PolicyBuilder struct {
 	// created policy object.
 	Object *policiesv1.Policy
 	// api client to interact with the cluster.
-	apiClient *clients.Settings
+	apiClient runtimeclient.Client
 	// used to store latest error message upon defining or mutating application definition.
 	errorMsg string
 }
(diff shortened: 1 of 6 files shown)


pkg/ocm/policy_test.go
```diff
@@ -278,6 +278,11 @@ func TestPolicyUpdate(t *testing.T) {
 			assert.Nil(t, err)
 		}
 
+		// This causes an error while updating and allows us to test the code path for force updates.
+		if testCase.force {
+			testBuilder.Definition.ResourceVersion = testBuilder.Definition.Name
```
kononovn, line 283: Why do you set ResourceVersion = Name ??

pkg/ocm/policy.go
```diff
@@ -198,30 +218,36 @@ func (builder *PolicyBuilder) Update(force bool) (*PolicyBuilder, error) {
...
 
 				return nil, err
 			}
 
 			return builder.Create()
 		}
-	}
-
-	if err == nil {
+	} else {
```
kononovn, line 250: I think you can simply:

```
return builder, err
```

kononovn (dismissed):
/lgtm

sebrandon1 (approved):

PR #555, opened 2024-07-31
ocm: removed kac and mcl types from clients

klaskosk:
This shifts the kac and managedcluster types to use scheme attachers in the functions themselves and removes them from the clients package.


kononovn (approved):
/lgtm

PR #557, opened 2024-07-31
olm: clusterserviceversion added unit tests

kononovn:


pkg/olm/clusterserviceversion.go
```diff
@@ -59,20 +69,44 @@ func PullClusterServiceVersion(apiClient *clients.Settings, name, namespace stri
...
+
+// Exists checks whether the given ClusterService exists.
 func (builder *ClusterServiceVersionBuilder) Exists() bool {
 	if valid, _ := builder.validate(); !valid {
 		return false
 	}
 
 	glog.V(100).Infof(
-		"Checking if clusterserviceversion %s exists",
+		"Checking if ClusterService %s exists",
```
klaskosk, line 105: Should this have been changed? It's still a ClusterServiceVersion, right?

pkg/olm/clusterserviceversion_test.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	for _, testCase := range testCases {
+		successful, err := testCase.clusterService.IsSuccessful()
+		if testCase.expectedError != nil {
+			assert.Equal(t, testCase.expectedError.Error(), err.Error())
+		} else {
+			assert.Nil(t, err)
+			assert.Equal(t, testCase.successful, successful)
+		}
```
klaskosk, line 274:
```suggestion
		assert.Equal(t, testCase.expectedError, err)
		
		if testCase.expectedError == nil {
			assert.Equal(t, testCase.successful, successful)
		}
```

pkg/olm/clusterserviceversion.go
```diff
@@ -4,13 +4,13 @@ import (
...
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-
 	"github.com/golang/glog"
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
 	"github.com/openshift-kni/eco-goinfra/pkg/msg"
-	oplmV1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	oplmV1alpha1 "github.com/openshift-kni/eco-goinfra/pkg/schemes/olm/operators/v1alpha1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
```
klaskosk, line 13: Since this PR uses testSchemes and the runtimeClient, should it also include switching `ClusterServiceVersionBuilder.apiClient` to be `runtimeClient.Client`? Or is that in a future PR?

kononovn: Due to the fact that I need to switch the rest of the objects under olm operator I decided to implement it later. I don's want to have huge PRs.

pkg/olm/clusterserviceversion_test.go
kononovn: The return error can be tricky and I need to convert it to string. The code that you proposed didn't work.

pkg/olm/clusterserviceversionlist.go
```diff
@@ -5,23 +5,25 @@ import (
...
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // ListClusterServiceVersion returns clusterserviceversion inventory in the given namespace.
 func ListClusterServiceVersion(
 	apiClient *clients.Settings,
 	nsname string,
-	options ...metav1.ListOptions) ([]*ClusterServiceVersionBuilder, error) {
+	options ...client.ListOptions) ([]*ClusterServiceVersionBuilder, error) {
 	if nsname == "" {
```
trewest, line 20: Also add:
```

	if apiClient == nil {
		glog.V(100).Infof("The apiClient cannot be nil")

		return nil, fmt.Errorf("clusterserviceversion 'apiClient' cannot be empty")
	}
```

trewest (approved):
/lgtm

PR #558, opened 2024-07-31
bmc: fix test failure in action

klaskosk:
Depending on how the timing works out, the BMC error message may be due to a context deadline exceeded or an i/o timeout. The latter is not accounted for in the test case and may cause errors during the test incoming changes action, like in [this test run].

This PR addresses it by asserting that either of these errors happened. Since both are due to timeouts, it does not change the expected behavior of the test.

[this test run]: https://github.com/klaskosk/eco-goinfra/actions/runs/10186752240/job/28179274304


kononovn (approved):
/lgtm

PR #573, opened 2024-08-07
sriov: SriovNetwork switched to runtime api client

kononovn:


pkg/sriov/network.go
```diff
@@ -437,9 +487,15 @@ func (builder *NetworkBuilder) Update(force bool) (*NetworkBuilder, error) {
...
+		return nil, fmt.Errorf("failed to update SriovNetwork, object does not exist on cluster")
+	}
+
+	err := builder.apiClient.Update(context.TODO(), builder.Definition)
+
+	if err == nil {
+		builder.Object = builder.Definition
+	}
 
 	if err != nil {
```
klaskosk, line 500: These are mutually exclusive so can just be an if-else instead of two ifs

klaskosk (approved):
/lgtm

PR #579, opened 2024-08-08
clients: update test client to support modifying test builder

trewest:
Gives user's the ability to customize their test builder within test cases when more advanced configuration is needed


kononovn (approved):
/lgtm

PR #588, opened 2024-08-09
bmh: add wait until annotation exists and tests

klaskosk:
This adds a method to wait until an annotation exists on the BMH along with unit tests for it.


trewest (approved):
/lgtm

PR #589, opened 2024-08-09
ocm: added wait for status message to contain to policy 

klaskosk:
WaitForMessageToContain waits until the latest message on the policy contains the expected message. Unit tests have also been added.


kononovn (approved):
/lgtm

PR #590, opened 2024-08-09
argocd: add sync config and internal schemes

klaskosk:
The application types are set to sync false since they require manual intervention to avoid bringing in too many dependencies.


trewest (approved):
/lgtm

PR #606, opened 2024-08-13
nmstate: nmstate added unit tests

kononovn:


pkg/nmstate/nmstate_test.go
```diff
@@ -1 +1,313 @@
...
+			assert.Equal(t, testNmState.Namespace, builderResult.Object.Namespace)
+		}
+	}
+}
+
+func TestNMStateNewBuilder(t *testing.T) {
+	testCases := []struct {
+		name              string
+		expectedErrorText string
+		client            bool
```
klaskosk, line 109: It doesn't look like the client field is ever used in the test. testSettings is always non-nil on line 126

pkg/nmstate/nmstate.go
```diff
@@ -137,7 +157,9 @@ func (builder *Builder) Update(force bool) (*Builder, error) {
 
 	err := builder.apiClient.Update(context.TODO(), builder.Definition)
 
-	if err != nil {
+	if err == nil {
+		builder.Object = builder.Definition
+	} else if err != nil {
```
trewest, line 162: Why else if?

kononovn: +1. fixed

pkg/nmstate/nmstate_test.go
```diff
@@ -1 +1,310 @@
...
+			name:              "",
+			expectedErrorText: "NMState 'name' cannot be empty",
+		},
+		{
+			name: "test1",
+		},
+	}
+	for _, testCase := range testCases {
+
+		testSettings := clients.GetTestClients(clients.TestClientParams{})
```
klaskosk, line 123: I think we should still test for when apiClient is nil, it just wasn't being tested before

klaskosk (approved):
/lgtm

trewest (approved):
/lgtm

PR #607, opened 2024-08-13
imageregistry: add wait for condition

klaskosk:
Like in the cgu package, this PR adds waiting for a condition to the imageregistry package. Unit tests for the function are included.


kononovn (approved):
/lgtm

PR #610, opened 2024-08-14
nmstate: added policy unit tests

kononovn:


pkg/nmstate/policy_test.go
```diff
@@ -0,0 +1,483 @@
...
+	testBuilder = buildValidPolicyTestBuilder(testSettings).WithOptions(
+		func(builder *PolicyBuilder) (*PolicyBuilder, error) {
+			return builder, fmt.Errorf("error")
+		})
+
+	assert.Equal(t, "error", testBuilder.errorMsg)
+}
+
+func TestPolicyWaitUntilCondition(t *testing.T) {
+	testCases := []struct {
```
klaskosk, line 429: I think there should also be a test case for when the policy does not exist
```go
{
	testNMStatePolicy: buildValidPolicyTestBuilder(clients.GetTestClients(clients.TestClientParams{})),
	expectedError:     fmt.Errorf("cannot wait for NodeNetworkConfigurationPolicy condition because it does not exist"),
	condition:         shared.NodeNetworkStateConditionAvailable,
},
```

klaskosk (approved):
/lgtm

PR #611, opened 2024-08-15
nmstate:list add u-test,removed scheme from client

kononovn:


pkg/nmstate/policylist.go
```diff
@@ -29,7 +42,7 @@ func ListPolicy(apiClient *clients.Settings, options ...goclient.ListOptions) ([
 	glog.V(100).Infof(logMessage)
 
 	policyList := &nmstateV1.NodeNetworkConfigurationPolicyList{}
-	err := apiClient.Client.List(context.TODO(), policyList, &passedOptions)
+	err = apiClient.Client.List(context.TODO(), policyList, &passedOptions)
 
 	if err != nil {
 		glog.V(100).Infof("Failed to list NodeNetworkConfigurationPolicy due to %s", err.Error())
```
klaskosk, line 48: Does the apiClient on line 58 need to be updated to `apiClient.Client`?

klaskosk (approved):
/lgtm

pkg/nmstate/policylist.go
kononovn: +1 fixed.

PR #614, opened 2024-08-15
ocm: switch mcl and kac to internal types

klaskosk:
This PR adds a config for the ocm package and adds both the managedcluster and klusterletaddonconfig types to the internal schemes. This allows removing the kac operator dependency and making the ocm api package an indirect depencency.


kononovn (approved):
/lgtm

PR #622, opened 2024-08-19
cgu: reset resource version in force update

klaskosk:
Similar to #483, when the CGU has ResourceVersion set on the definition, it can cause the create to fail. This results in the CGU being deleted but not created by the force update.

This situation can be achieved by calling WaitForCondition on a CGU before a force update while the CGU gets updated on the cluster in between. Has appeared in the TALM tests.


kononovn (approved):
/lgtm

PR #627, opened 2024-08-20
kmm: module added unit tests and runtime client

kononovn:


pkg/kmm/module_test.go
```diff
@@ -0,0 +1,561 @@
...
+var (
+	testSchemesV1beta1 = []clients.SchemeAttacher{
+		v1beta1.AddToScheme,
+	}
+	defaultModuleName      = "testmodule"
+	defaultModuleNamespace = "testns"
+)
+
+func TestNewModuleBuilder(t *testing.T) {
+	testCases := []struct {
```
klaskosk, line 24: should also have a test case to test for a nil client

klaskosk (approved):
/lgtm

PR #629, opened 2024-08-21
mco: tests and runtime client for machine config

klaskosk:
* Switches the MachineConfig resource to use the runtime client and adds unit tests for the file.
* Removes the MCO client from the test clients. Note that it is still in clients.Settings until MCP is converted to runtime client.
* Renames a variable in the KubeletConfig tests that had been copied from the ocm package.


kononovn (approved):
/lgtm

PR #631, opened 2024-08-22
mco: unit tests and runtime client for mcp

klaskosk:
This switches the MCP resource to the runtime client and adds unit tests for most of the methods on the builder.

For brevity, the unit tests for the list functions and the wait*** methods will come in #632.


kononovn (approved):
/lgtm

PR #632, opened 2024-08-22
mco: unit tests and runtime client for mcp list 

klaskosk:
Added unit tests and switched to the runtime client for the MCP list functions. Also adds unit tests to the remaining untested MCP functions. Now that the package is complete, removes the machineconfig client from clients.


kononovn (approved):
/lgtm

PR #634, opened 2024-08-26
argocd: add application wait for condition and tests

klaskosk:
Like in the imageregistry and cgu packages, this PR adds a method for waiting for a condition to occur on the object, along with the corresponding unit tests.


kononovn (approved):
/lgtm

PR #638, opened 2024-08-29
deployment: remove pointer from containerSpec

sebrandon1:
To match the `daemonset` line of the same operation.  The `daemonset` package doesn't require the containerSpec to be a pointer so we can remove it here to prevent `nil` dereferencing.


sebrandon1:
This will take changes in the eco-gotests repo as well.

kononovn:
@sebrandon1 This is critical api change. What is the root cause behind that?

PR #639, opened 2024-08-29
Adjust deployment pkg return consistency

sebrandon1:

### DIFF
--- a/pkg/deployment/deployment.go
+++ b/pkg/deployment/deployment.go
@@ -69,18 +69,24 @@ func NewBuilder(
 		glog.V(100).Infof("The name of the deployment is empty")
 
 		builder.errorMsg = "deployment 'name' cannot be empty"
+
+		return &builder
 	}
 
 	if nsname == "" {
 		glog.V(100).Infof("The namespace of the deployment is empty")
 
 		builder.errorMsg = "deployment 'namespace' cannot be empty"
+
+		return &builder
 	}
 
 	if len(labels) == 0 {
 		glog.V(100).Infof("There are no labels for the deployment")
 
 		builder.errorMsg = "deployment 'labels' cannot be empty"
+
+		return &builder
 	}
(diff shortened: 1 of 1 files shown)


pkg/deployment/deployment.go
```diff
@@ -69,18 +69,24 @@ func NewBuilder(
 		glog.V(100).Infof("The name of the deployment is empty")
 
 		builder.errorMsg = "deployment 'name' cannot be empty"
+
+		return &builder
```
kononovn, line 73: No need for this repeated return. We have validate method which will print that builder has error and return the exact error.

pkg/deployment/deployment.go
```diff
@@ -451,7 +457,7 @@ func (builder *Builder) CreateAndWaitUntilReady(timeout time.Duration) (*Builder
 	if _, err := builder.Create(); err != nil {
 		glog.V(100).Infof("Failed to create deployment. Error is: '%s'", err.Error())
 
-		return nil, fmt.Errorf(err.Error())
+		return nil, err
```
kononovn, line 460: +1

kononovn (changes requested):
1 minor comment

pkg/deployment/deployment.go
kononovn: We use builder.errorMsg in our unit test. You just need to check that builder has relevant error.

sebrandon1: I think we've talked about this before for other parts of the code where you could potentially have a blank string for name, namespace and an empty `if len(labels) == 0 {` and the last condition would be the builder.errorMsg that was set.

kononovn: yes and It's OK because code will not work until all errors I'll be fixed.

kononovn (approved):
/lgtm

PR #648, opened 2024-09-06
pod: add basic unit tests

klaskosk:
This PR adds unit tests to the basic pod functions (NewBuilder, Pull, Create, Delete, Get, Exists) and refactors the existing tests in the package to be more in line with those from other packages.

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -64,19 +64,25 @@ func NewBuilder(apiClient *clients.Settings, name, nsname, image string) *Builde
 	if name == "" {
 		glog.V(100).Infof("The name of the pod is empty")
 
-		builder.errorMsg = "pod's name is empty"
+		builder.errorMsg = "pod 'name' cannot be empty"
+
+		return builder
 	}
 
 	if nsname == "" {
 		glog.V(100).Infof("The namespace of the pod is empty")
 
-		builder.errorMsg = "namespace's name is empty"
+		builder.errorMsg = "pod 'namespace' cannot be empty"
+
+		return builder
 	}
 
 	if image == "" {
 		glog.V(100).Infof("The image of the pod is empty")
 
-		builder.errorMsg = "pod's image is empty"
+		builder.errorMsg = "pod 'image' cannot be empty"
+
+		return builder
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/pod/pod.go
```diff
@@ -169,6 +173,27 @@ func (builder *Builder) DefineOnNode(nodeName string) *Builder {
 	return builder
 }
 
+// Get returns a Pod object if found.
+func (builder *Builder) Get() (*corev1.Pod, error) {
```
kononovn, line 177: Pod is standard package so we don't need Get function because Exist function provides this functionality. You can remove Get

pkg/pod/pod_test.go
```diff
@@ -1,242 +1,463 @@
...
-		"test-container",
-		"registry.example.com/test/image:latest",
-		[]string{"/bin/sh", "-c", "sleep infinity"})
+const (
+	defaultPodName   = "test-pod"
+	defaultPodNsName = "test-ns"
+	defaultPodImage  = "test-image"
+)
+
+func TestNewBuilder(t *testing.T) {
```
kononovn, line 22: Please add TestPod prefix to all test cases.

kononovn (changes requested):
Just few comments to fix

pkg/pod/pod_test.go
```diff
@@ -1,242 +1,430 @@
...
+			testBuilder:   buildValidPodTestBuilder(buildTestClientWithDummyPod()),
+			expectedError: nil,
+		},
+		{
+			testBuilder:   buildInvalidPodTestBuilder(buildTestClientWithDummyPod()),
+			expectedError: fmt.Errorf("pod 'namespace' cannot be empty"),
+		},
+		{
+			testBuilder:   buildValidPodTestBuilder(clients.GetTestClients(clients.TestClientParams{})),
+			expectedError: fmt.Errorf("pod cannot be deleted because it does not exist"),
```
kononovn, line 210: Please change function delete. We should print log but don't return error.

kononovn (changes requested):
Just few minor comments.

kononovn (approved):
/lgtm

PR #652, opened 2024-09-10
reporter: remove reporter package

klaskosk:
This is an older version of the eco-gotests reporter package, but eco-gotests does not even use it so it is dead code currently.

Using the eco-gotests version exclusively also allows us to add all the schemes in that package without including the dependencies in eco-goinfra.


pkg/reporter/reporter.go
```diff
@@ -1,129 +0,0 @@
...
-	res, err := k8sreporter.New("", apiScheme, nsToDumpFilter, reportPath, cRDs...)
-
-	if err != nil {
-		return nil, err
-	}
-
-	return res, nil
-}
-
-// ReportIfFailed dumps requested cluster CRs if TC is failed to the given directory.
```
achuzhoy, line 48: nit: // ReportIfFailed dumps requested cluster CRs to the given directory if TC is failed.

trewest (approved):
/lgtm

pkg/reporter/reporter.go
klaskosk: @achuzhoy This code is being deleted in this PR, so there are no changes to be made here

PR #657, opened 2024-09-11
bmh: add HostFirmwareSettings resource and unit tests

klaskosk:
Adds the HostFirmwareSettings resource to the bmh package and unit tests for the functions it exports.

### DIFF
--- a/pkg/bmh/hfs.go
+++ b/pkg/bmh/hfs.go
@@ -1,0 +1,196 @@
+package bmh
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	bmhv1alpha1 "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// HFSBuilder provides a struct to interface with HostFirmwareSettings resources on a specific cluster.
+type HFSBuilder struct {
+	// Definition of the HostFirmwareSettings used to create the object.
+	Definition *bmhv1alpha1.HostFirmwareSettings
+	// Object of the HostFirmwareSettings as it is on the cluster.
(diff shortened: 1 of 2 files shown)


klaskosk:
cc: @dgonyier

pkg/bmh/hfs.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// HFSBuilder provides a struct to interface with HostFirmwareSettings resources on a specific cluster.
+type HFSBuilder struct {
+	Definition *bmhv1alpha1.HostFirmwareSettings
```
kononovn, line 18: Please add comment to public element

pkg/bmh/hfs.go
```diff
@@ -0,0 +1,194 @@
...
+	Object     *bmhv1alpha1.HostFirmwareSettings
```
kononovn, line 19: Please add comment to public element

pkg/bmh/hfs_test.go
```diff
@@ -0,0 +1,231 @@
...
+			assert.Nil(t, err)
+			assert.Equal(t, testHFS.Name, hfsBuilder.Definition.Name)
+			assert.Equal(t, testHFS.Namespace, hfsBuilder.Definition.Namespace)
+		} else {
+			assert.EqualError(t, err, testCase.expectedErrorText)
+		}
+	}
+}
+
+func TestHFSGet(t *testing.T) {
```
kononovn, line 96: I think 1 scenario is missing for all test cases: 
buildValidHFSBuilder(clients.GetTestClients(clients.TestClientParams{})),

kononovn (changes requested):
Please fix few comments

pkg/bmh/hfs_test.go
klaskosk: @kononovn I use the `buildTestClientWithHFSScheme` function instead since if I don't add the scheme to the client I'll get `"no kind is registered for the type v1alpha1.HostFirmwareSettings in scheme \"pkg/clients/clients.go:302\""` errors instead of the expected. So it's equivalent, just adding the scheme even with no runtime objects

kononovn (approved):
/lgtm

PR #659, opened 2024-09-12
rbac: added clusterrole unit tests

kononovn:


pkg/rbac/clusterrole.go
```diff
@@ -232,7 +241,7 @@ func (builder *ClusterRoleBuilder) Exists() bool {
 // validate will check that the builder and builder definition are properly initialized before
 // accessing any member fields.
 func (builder *ClusterRoleBuilder) validate() (bool, error) {
-	resourceCRD := "ClusterRole"
+	resourceCRD := "clusterRole"
```
achuzhoy, line 244: Why? new format?
I see plenty of examples where it's capitalized in the repo.

pkg/rbac/clusterrole_test.go
```diff
@@ -0,0 +1,354 @@
...
+func TestNewClusterRoleBuilder(t *testing.T) {
+	testCases := []struct {
+		name              string
+		nsName            string
+		rule              v1.PolicyRule
+		expectedErrorText string
+		client            bool
+	}{
+		{
+			name: "test",
```
achuzhoy, line 24: nsName not set. On purpose?

pkg/rbac/clusterrole_test.go
```diff
@@ -0,0 +1,354 @@
...
+
+	assert.Equal(t, "", testBuilder.errorMsg)
+	testBuilder = buildValidRoleBuilder(testSettings).WithOptions(
+		func(builder *RoleBuilder) (*RoleBuilder, error) {
+			return builder, fmt.Errorf("error")
+		})
+	assert.Equal(t, "error", testBuilder.errorMsg)
+}
+
+// buildValidTestBuilder returns a valid Builder for testing purposes.
```
achuzhoy, line 329:
```suggestion
// buildValidClusterRoleBuilder returns a valid Builder for testing purposes.
```

pkg/rbac/clusterrole_test.go
```diff
@@ -0,0 +1,354 @@
...
+	testCases := []struct {
+		testClusterRole *ClusterRoleBuilder
+		expectedError   error
+	}{
+		{
+			testClusterRole: buildValidClusterRoleBuilder(buildTestClientWithClusterRoleDummyObject()),
+			expectedError:   nil,
+		},
+		{
+			testClusterRole: buildInvalidClusterRoleTestBuilder(buildTestClientWithClusterRoleDummyObject()),
```
achuzhoy, line 216: do we need this one?

pkg/rbac/clusterrole.go
```diff
@@ -44,7 +44,6 @@ func NewClusterRoleBuilder(apiClient *clients.Settings, name string, rule v1.Pol
 			ObjectMeta: metav1.ObjectMeta{
 				Name: name,
 			},
-			Rules: []v1.PolicyRule{rule},
```
trewest, line 47: Why does it need to be removed?

sebrandon1: I agree this should be removed.  When I was trying to just run a simple `Delete()` in my project I had to populate a rule prior to issuing a delete.

sebrandon1:
Can we remove the parameter for the `rule` in the `NewClusterRoleBuilder` func?  Potentially just create the builder then have an additional func that accepts a rule to add to the object.Definition?

pkg/rbac/clusterrole.go
kononovn: Yes and this is an issue. We print it in the begging of the error message. In go all error has to start from small letter. Capitalized letters has to be avoided.

pkg/rbac/clusterrole_test.go
kononovn: clusterRole is not namespaces resource.

pkg/rbac/clusterrole_test.go
kononovn: Yes it simulates error in builder.

pkg/rbac/clusterrole.go
kononovn: @trewest It's  bug that i've found in unit test. We used to add Rules 2 times. 
First time on line 47 and second time on the next line using `builder.WithRules([]v1.PolicyRule{rule})` . So as a result we had 2 identical rules

kononovn:
> Can we remove the parameter for the `rule` in the `NewClusterRoleBuilder` func? Potentially just create the builder then have an additional func that accepts a rule to add to the object.Definition?

This is public function and it's not good idea to change public functions because it'll require to adjust other projects.  Let's try to keep Public API as static as possible.

pkg/rbac/clusterrole_test.go
trewest: Can you remove nsName from the inputs of the test?

sebrandon1:
@kononovn Could we add a public API function to add the rule, then change the eco-tests references to add the rules separately, then come back later and remove the `rule` parameter when the dependent func calls are changed?  Staggered approach.

kononovn:
> @kononovn Could we add a public API function to add the rule, then change the eco-tests references to add the rules separately, then come back later and remove the `rule` parameter when the dependent func calls are changed? Staggered approach.

Yes we can, however we need to open an issue and discuss it during our bi-weekly. If we have agreement then we'll fix it int the next release.

pkg/rbac/clusterrole_test.go
kononovn: fixed

trewest (approved):
/lgtm

PR #664, opened 2024-09-13
nodes: add unit tests for all but drain

klaskosk:
Added unit tests to all of the node functions except the ones that involve draining and the drain helper.


kononovn (approved):
/lgtm

PR #681, opened 2024-09-19
Ensure Delete() funcs use Definition

sebrandon1:
Be consistent in the `Delete()` funcs in that they should be using the `builder.Definition` instead of the `builder.Object` structs when referencing variables.  It's a mix right now and I feel that the `Definition` should be where we are gathering that information.


klaskosk (approved):
/lgtm

kononovn (approved):
/lgtm

PR #672, opened 2024-09-17
statefulset: add Delete() func

sebrandon1:
Adds the functionality and the unit test for a Delete() func in the statefulset package.

PR #687, opened 2024-09-23
storageclass: add glog to Delete(); return improvements

sebrandon1:
Builds on changes from: #379 

As noted in #672 we should have a `glog` entry for when an object does not exist.


pkg/storage/storageclass.go
```diff
@@ -370,13 +370,13 @@ func (builder *ClassBuilder) validate() (bool, error) {
...
 
-		builder.errorMsg = msg.UndefinedCrdObjectErrString(resourceCRD)
+		return false, fmt.Errorf(msg.UndefinedCrdObjectErrString(resourceCRD))
 	}
 
 	if builder.apiClient == nil {
 		glog.V(100).Infof("The %s builder apiclient is nil", resourceCRD)
 
-		builder.errorMsg = fmt.Sprintf("%s builder cannot have nil apiClient", resourceCRD)
+		return false, fmt.Errorf("%s builder cannot have nil apiClient", resourceCRD)
```
klaskosk, line 379: resourceCRD should be updated to `storageClass` so that the error starts with a lowercase

sebrandon1: Done.

klaskosk (approved):
/lgtm

klaskosk:
approved too soon again lol, but just change to lowercase in the other areas and it's fine

klaskosk (approved):

pkg/storage/storageclass.go
```diff
@@ -79,9 +83,7 @@ func (builder *ClassBuilder) WithReclaimPolicy(
 		glog.V(100).Infof("The reclaimPolicy of the storageclass is empty")
 
 		builder.errorMsg = "storageclass 'reclaimPolicy' cannot be empty"
-	}
 
-	if builder.errorMsg != "" {
```
kononovn, line 84: If you remove this condition we add invalid reclaimPolicy to builder. Please revert it back

pkg/storage/storageclass.go
```diff
@@ -101,9 +103,7 @@ func (builder *ClassBuilder) WithVolumeBindingMode(
 		glog.V(100).Infof("The bindingMode of the storageclass is empty")
 
 		builder.errorMsg = "storageclass 'bindingMode' cannot be empty"
-	}
 
-	if builder.errorMsg != "" {
```
kononovn, line 106: If you remove this condition we add invalid bindingMode to builder. Please revert it back

kononovn (changes requested):
Please check comments

pkg/storage/storageclass.go
sebrandon1: The diff here on Github makes it look like that would happen.  It makes it look confusing when its not.

image

pkg/storage/storageclass.go
sebrandon1: Same here, the diff on Github makes it look confusing for some reason.  The actual code won't allow an invalid bindingMode though.

image

pkg/storage/storageclass.go
```diff
@@ -246,6 +246,10 @@ func (builder *ClassBuilder) Delete() error {
 	glog.V(100).Infof("Deleting storageclass %s", builder.Definition.Name)
 
 	if !builder.Exists() {
+		glog.V(100).Infof("storageclass %s does not exist", builder.Definition.Name)
+
+		builder = nil
```
kononovn, line 251: Please fix it. builder.Object = nil

kononovn (changes requested):
1 comment to fix

pkg/storage/storageclass.go
sebrandon1: Done!

kononovn (approved):
/lgtm

PR #690, opened 2024-09-23
baremetalhost: Delete() should return nil if not exists

sebrandon1:
Similar to other Delete() functions, we do not return errors when an object already doesn't exist.

Builds from changes in: #379


klaskosk (approved):
/lgtm

kononovn (approved):
/lgtm

PR #693, opened 2024-09-24
siteconfig: add internal types to schemes

klaskosk:
Since the siteconfig types import hive and assisted, both of which have internal types in schemes, this adds the siteconfig types and replaces the hive and assisted imports to go to our internal schemes.

cc: @mpmaruthu


kononovn (approved):
/lgtm

PR #694, opened 2024-09-24
rbac: v1 to rbacv1 import consistency

sebrandon1:
Similar to: https://github.com/openshift-kni/eco-goinfra/pull/387

Just changing `v1` to `rbacv1` to better match the rest of the repo.


kononovn (approved):
/lgtm

PR #695, opened 2024-09-24
Assert: Adjust expected/actual parameter order

sebrandon1:
I noticed a number of spots where `assert.Equal` was being used incorrectly.  The order is `expected, actual` instead of `actual, expected` for the values of the parameters.


kononovn (approved):
/lgtm

PR #696, opened 2024-09-24
assisted/infraenv: add glog and return nil to Delete()

sebrandon1:
Built from: #379 

Adds a `glog` entry and returns nil instead of an error for Delete() that does not exist.

Similar to: #690


kononovn (approved):
/lgtm

PR #701, opened 2024-09-26
resourcequotas: add pkg and unit tests

sebrandon1:
Adds a `resourcequotas` package to eco-goinfra.

### DIFF
--- a/pkg/resourcequotas/list.go
+++ b/pkg/resourcequotas/list.go
@@ -1,0 +1,58 @@
+package resourcequotas
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// List returns resource quota inventory in the given namespace.
+func List(apiClient *clients.Settings, nsname string, options ...metav1.ListOptions) ([]*Builder, error) {
+	if nsname == "" {
+		glog.V(100).Infof("resource quota 'nsname' parameter can not be empty")
+
+		return nil, fmt.Errorf("failed to list resource quotas, 'nsname' parameter is empty")
+	}
+
+	passedOptions := metav1.ListOptions{}
+	logMessage := fmt.Sprintf("Listing resource quotas in the namespace %s", nsname)
+
+	if len(options) > 1 {
+		glog.V(100).Infof("'options' parameter must be empty or single-valued")
+
+		return nil, fmt.Errorf("error: more than one ListOptions was passed")
+	}
+
+	if len(options) == 1 {
+		passedOptions = options[0]
+		logMessage += fmt.Sprintf(" with the options %v", passedOptions)
+	}
+
+	glog.V(100).Infof(logMessage)
+
+	resourceQuotaList, err := apiClient.ResourceQuotas(nsname).List(context.TODO(), passedOptions)
+
+	if err != nil {
+		glog.V(100).Infof("Failed to list resource quotas in the namespace %s due to %s", nsname, err.Error())
+
+		return nil, err
+	}
+
+	var resourceQuotaObjects []*Builder
+
+	for _, runningResourceQuota := range resourceQuotaList.Items {
+		copiedResourceQuota := runningResourceQuota
+		resourceQuotaBuilder := &Builder{
+			apiClient:  apiClient.CoreV1Interface,
+			Object:     &copiedResourceQuota,
+			Definition: &copiedResourceQuota,
+		}
+
+		resourceQuotaObjects = append(resourceQuotaObjects, resourceQuotaBuilder)
+	}
+
+	return resourceQuotaObjects, nil
+}
--- a/pkg/resourcequotas/list.go
+++ b/pkg/resourcequotas/list.go
@@ -11,6 +11,8 @@ import (
 
 // List returns resource quota inventory in the given namespace.
 func List(apiClient *clients.Settings, nsname string, options ...metav1.ListOptions) ([]*Builder, error) {
+	glog.V(100).Infof("Listing resource quotas in the namespace %s", nsname)
+
 	if nsname == "" {
 		glog.V(100).Infof("resource quota 'nsname' parameter can not be empty")
 
--- a/pkg/resourcequotas/resourcequotas.go
+++ b/pkg/resourcequotas/resourcequotas.go
@@ -67,17 +67,6 @@ func NewBuilder(apiClient *clients.Settings, name, nsname string) *Builder {
 	return builder
 }
 
-// WithQuotaSpec sets the resource quota spec.
-func (builder *Builder) WithQuotaSpec(quotaSpec corev1.ResourceQuotaSpec) *Builder {
-	if valid, _ := builder.validate(); !valid {
-		return builder
-	}
-
-	builder.Definition.Spec = quotaSpec
-
-	return builder
-}
-
 // Pull retrieves the resource quota from the cluster.
 func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
 	if apiClient == nil {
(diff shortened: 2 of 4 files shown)


pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,198 @@
...
+	Object *corev1.ResourceQuota
+
+	// Used in functions that define or mutate deployment definition. errorMsg is processed before the deployment
+	// object is created.
+	errorMsg  string
+	apiClient corev1Typed.CoreV1Interface
+}
+
+// NewBuilder creates a new resource quota builder.
+func NewBuilder(apiClient corev1Typed.CoreV1Interface, name, nsname string,
```
klaskosk, line 31: is there a reason the entire definition needs to be provided when creating a new builder? It also means that the name and nsname parameters do nothing

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,198 @@
...
+
+	builder.Object, err = builder.apiClient.ResourceQuotas(
+		builder.Definition.Namespace).Get(context.TODO(),
+		builder.Definition.Name, metav1.GetOptions{})
+
+	return err == nil || !k8serrors.IsNotFound(err)
+}
+
+// validate validates the resource quota definition.
+func (builder *Builder) validate() (bool, error) {
```
klaskosk, line 121: unexported validate function should go after the exported functions

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,198 @@
...
+			name, nsname)
+	}
+
+	builder.Definition = builder.Object
+
+	return builder, nil
+}
+
+// Exists checks if the resource quota exists in the cluster.
+func (builder *Builder) Exists() bool {
```
klaskosk, line 103: There's no corresponding unit test for exists

pkg/resourcequotas/list.go
```diff
@@ -0,0 +1,58 @@
...
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// List returns resource quota inventory in the given namespace.
+func List(apiClient *clients.Settings, nsname string, options ...metav1.ListOptions) ([]*Builder, error) {
```
klaskosk, line 13: There's no corresponding unit test for list

pkg/resourcequotas/resourcequotas.go
klaskosk, line 31: I think for consistency it makes more sense to take `*clients.Settings` and then take the `CoreV1Interface` from it for the builder

sebrandon1: You're right.  I was working with my existing code in my QE repo and there we define it completely before creating.  So I think what I'll do is create a base `ResourceQuota` object and do a `WithXYZ` func to add additional spec options.

pkg/resourcequotas/resourcequotas.go
sebrandon1: Done.

pkg/resourcequotas/list.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas.go
sebrandon1: Done.

sebrandon1: Done.

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,214 @@
...
+	apiClient corev1Typed.CoreV1Interface
+}
+
+// NewBuilder creates a new resource quota builder.
+func NewBuilder(apiClient *clients.Settings, name, nsname string) *Builder {
+	glog.V(100).Infof("Initializing new resource quota structure with the following params: "+
+		"name=%s, namespace=%s", name, nsname)
+
+	builder := &Builder{
+		apiClient: apiClient.CoreV1Interface,
```
klaskosk, line 36: I think we still need to set the definition with ObjectMeta, otherwise name and nsname aren't used and Definition ends up nil. It seems like Definition being non-nil is an invariant in the builders

pkg/resourcequotas/resourcequotas.go
klaskosk: @sebrandon1 It doesn't look like this change is in the PR, does it still need to be pushed?

pkg/resourcequotas/resourcequotas_test.go
```diff
@@ -0,0 +1,361 @@
...
+			assert.Equal(t, resource.MustParse("1"), testBuilder.Definition.Spec.Hard[corev1.ResourceCPU])
+			assert.Equal(t, resource.MustParse("1Gi"), testBuilder.Definition.Spec.Hard[corev1.ResourceMemory])
+		} else {
+			assert.Nil(t, result)
+		}
+	}
+}
+
+func TestResourceQuotaCreate(t *testing.T) {
+	generateResourceQuota := func(name, namespace string) *corev1.ResourceQuota {
```
klaskosk, line 171: Can be moved into a function instead of a closure since it's used in multiple different functions

pkg/resourcequotas/resourcequotas.go
sebrandon1: Whoops yes, I missed changing this to the struct definition.

pkg/resourcequotas/resourcequotas_test.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,226 @@
...
+		glog.V(100).Info("Builder is nil")
+
+		return nil
+	}
+
+	if builder.Definition == nil {
+		glog.V(100).Info("Resource Quota definition is nil")
+
+		return nil
+	}
```
klaskosk, line 82:
```suggestion
	if valid, _ := builder.validate(); !valid {
		return builder
	}
```

klaskosk (commented):
Thanks for making the changes, I missed one thing but otherwise lgtm

pkg/resourcequotas/resourcequotas.go
sebrandon1: Done!

pkg/resourcequotas/list.go
```diff
@@ -0,0 +1,58 @@
...
+	if nsname == "" {
```
kononovn, line 14: Please add glog

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,218 @@
...
+
+	return builder
+}
+
+// WithQuotaSpec sets the resource quota spec.
+func (builder *Builder) WithQuotaSpec(quotaSpec corev1.ResourceQuotaSpec) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
```
kononovn, line 75: Please add glog

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,218 @@
...
+		builder.errorMsg = "resource quota 'namespace' cannot be empty"
+
+		return builder
+	}
+
+	return builder
+}
+
+// WithQuotaSpec sets the resource quota spec.
+func (builder *Builder) WithQuotaSpec(quotaSpec corev1.ResourceQuotaSpec) *Builder {
```
kononovn, line 71: Move this function down after Pull

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,218 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	builder.Object = nil
+
+	return nil
+}
+
```
kononovn, line 188: Can you add Update function?

pkg/resourcequotas/resourcequotas_test.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+
+		testBuilder := buildTestBuilderWithFakeObjects(runtimeObjects, testRQ.Name, testRQ.Namespace)
+		result := testBuilder.Exists()
+
+		assert.Equal(t, testCase.rqExistsAlready, result)
+	}
+}
+
+func TestNewBuilder(t *testing.T) {
```
kononovn, line 252: Please move this function of top of the file and add TestResourceQuota* Prefix

pkg/resourcequotas/resourcequotas_test.go
```diff
@@ -0,0 +1,404 @@
...
+				assert.NotNil(t, builderResult)
+				assert.Equal(t, testCase.expectedBuilder.Definition.Name, builderResult.Definition.Name)
+				assert.Equal(t, testCase.expectedBuilder.Definition.Namespace, builderResult.Definition.Namespace)
+				assert.Equal(t, testCase.expectedError, builderResult.errorMsg)
+			}
+		}
+	}
+}
+
+func TestValidate(t *testing.T) {
```
kononovn, line 319: Please add TestResourceQuota* Prefix

kononovn (changes requested):
Please fix few comments

pkg/resourcequotas/list.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas_test.go
sebrandon1: Done.

pkg/resourcequotas/resourcequotas_test.go
sebrandon1: Done.

klaskosk (approved):
/lgtm

klaskosk (changes requested):

pkg/resourcequotas/resourcequotas.go
```diff
@@ -0,0 +1,261 @@
...
+			err := builder.Delete()
+			if err != nil {
+				glog.V(100).Infof(msg.FailToUpdateError("resource quota",
+					builder.Definition.Name, builder.Definition.Namespace))
+			}
+
+			return nil, err
+		}
+
+		return builder.Create()
```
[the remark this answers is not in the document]
klaskosk: Well the `if err != nil` block also needs a `return nil, err` but the more important part is making sure `return builder.Create()` is inside the `if force` block

kononovn (approved):
/lgtm

pkg/resourcequotas/resourcequotas.go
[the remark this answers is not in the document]
sebrandon1: Fixed the return!

PR #705, opened 2024-09-28
siteconfig: Add 'NewBuilder, Update, WaitForCondition...' and other methods

mpmaruthu:
Addding 'NewClusterInstanceBuilder, Update, WaitFor...Condition' methods in siteconfig package.

@klaskosk , @trewest , Please help to review this PR when you get a time. Thanks!

### DIFF
--- a/pkg/siteconfig/clusterinstance.go
+++ b/pkg/siteconfig/clusterinstance.go
@@ -3,18 +3,23 @@ package siteconfig
 import (
 	"context"
 	"fmt"
+	"strings"
+	"time"
 
 	"github.com/golang/glog"
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
 	"github.com/openshift-kni/eco-goinfra/pkg/msg"
 	siteconfigv1alpha1 "github.com/openshift-kni/eco-goinfra/pkg/schemes/siteconfig/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/apimachinery/pkg/util/wait"
 	goclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ClusterInstanceBuilder provides struct for the ClusterInstance object.
-type ClusterInstanceBuilder struct {
+// CIBuilder provides struct for the ClusterInstance object.
+type CIBuilder struct {
 	// ClusterInstance definition. Used to create a clusterinstance object.
 	Definition *siteconfigv1alpha1.ClusterInstance
 	// Created clusterinstance object.
@@ -26,8 +31,56 @@ type ClusterInstanceBuilder struct {
 	errorMsg string
 }
 
+// NewCIBuilder creates a new instance of CIBuilder.
+func NewCIBuilder(apiClient *clients.Settings, name, nsname string) *CIBuilder {
+	glog.V(100).Infof(
+		"Initializing new ClusterInstance structure with the following params: name: %s, nsname: %s",
+		name, nsname)
+
+	if apiClient == nil {
+		glog.V(100).Info("The apiClient for the clusterinstance is nil")
+
+		return nil
+	}
+
+	err := apiClient.AttachScheme(siteconfigv1alpha1.AddToScheme)
+	if err != nil {
+		glog.V(100).Infof("Failed to add siteconfig v1alpha1 scheme to client schemes")
+
+		return nil
+	}
+
+	builder := &CIBuilder{
+		apiClient: apiClient.Client,
+		Definition: &siteconfigv1alpha1.ClusterInstance{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: nsname,
+			},
+		},
+	}
+
+	if name == "" {
+		glog.V(100).Infof("The name of the clusterinstance is empty")
+
+		builder.errorMsg = "clusterinstance 'name' cannot be empty"
+
+		return builder
+	}
+
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the clusterinstance is empty")
+
+		builder.errorMsg = "clusterinstance 'nsname' cannot be empty"
+
+		return builder
+	}
+
+	return builder
+}
+
 // PullClusterInstance retrieves an existing ClusterInstance from the cluster.
-func PullClusterInstance(apiClient *clients.Settings, name, nsname string) (*ClusterInstanceBuilder, error) {
+func PullClusterInstance(apiClient *clients.Settings, name, nsname string) (*CIBuilder, error) {
 	glog.V(100).Infof(
 		"Pulling existing clusterinstance with name %s from namespace %s", name, nsname)
 
@@ -45,7 +98,7 @@ func PullClusterInstance(apiClient *clients.Settings, name, nsname string) (*Clu
 		return nil, fmt.Errorf("failed to add siteconfigv1alpha1 to client schemes")
 	}
 
-	builder := &ClusterInstanceBuilder{
+	builder := &CIBuilder{
 		apiClient: apiClient.Client,
 		Definition: &siteconfigv1alpha1.ClusterInstance{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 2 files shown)


pkg/siteconfig/clusterinstance.go
```diff
@@ -162,9 +268,233 @@ func (builder *ClusterInstanceBuilder) Exists() bool {
 	return err == nil || !k8serrors.IsNotFound(err)
 }
 
+// WaitForCIVCondition waits until the ClusterInstance type:ClusterInstanceValidated
+// has a condition that matches the expected, checking only the Type, Status, Reason, and Message fields.
+// For the message field, it matches if the message contains the expected.
+// Zero fields in the expected condition are ignored.
+func (builder *CIBuilder) WaitForCIVCondition(expected metav1.Condition, timeout time.Duration) (*CIBuilder, error) {
```
klaskosk, line 275: @mpmaruthu It looks like the actual type of the condition you're waiting for changes in the name but not the code for the function. Could you just have one `WaitForCondition` function and then use conditions as variables?

pkg/siteconfig/clusterinstance.go
```diff
@@ -118,8 +181,51 @@ func (builder *ClusterInstanceBuilder) Create() (*ClusterInstanceBuilder, error)
...
+	}
+
+	err := builder.apiClient.Update(context.TODO(), builder.Definition)
+
+	if err != nil {
+		if force {
+			glog.V(100).Infof(
+				msg.FailToUpdateNotification("clusterinstance", builder.Definition.Name, builder.Definition.Namespace))
+
+			err := builder.Delete()
```
klaskosk, line 207:
```suggestion
			err := builder.Delete()
			builder.Definition.ResourceVersion = ""
```
There's an edge case that can cause this function to fail if you don't reset the resource version. Sometimes it's fine but I like to reset it anyway to prevent issues later

pkg/siteconfig/clusterinstance.go
```diff
@@ -11,10 +13,23 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+var (
```
klaskosk, line 20: I think I'd prefer the conditions being in the test suite. My reasoning is that conditions may change between releases (like cgu in 4.12)  so older ones are only available with older release branches of eco-goinfra. But cnf/ran always uses the latest version of eco-goinfra so we'd have to create our own variables for older conditions. If we're creating our own variables in the test suite then might as well only do that. Let me know if you disagree though I'm open to hearing alternatives

kononovn:
@mpmaruthu Please re-base

pkg/siteconfig/clusterinstance.go
mpmaruthu: Good point, I do agree the proposed approach to declare "ClusterInstanceConditionType" as variables in test suite instead of declaring it in the API/package(siteconfig) code.

pkg/siteconfig/clusterinstance.go
mpmaruthu: Okay, updated. Thanks! The proposed change will be in place after submitting a new patch for review.

kononovn:
@klaskosk Can you review?

pkg/siteconfig/clusterinstance.go
mpmaruthu: Updated. Thanks!

pkg/siteconfig/clusterinstance.go
mpmaruthu: Taken care. changes are updated accordingly.

pkg/siteconfig/clusterinstance_test.go
```diff
@@ -13,6 +15,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+var (
+	defaultClusterInstanceCondition = ClusterInstanceCondition
```
klaskosk, line 19:
```suggestion
	defaultClusterInstanceCondition = metav1.Condition{Type: "", Status: metav1.ConditionTrue}
```
and then could you please remove the variable from `clusterinstance.go`

mpmaruthu: Okay, updated. Thanks! The proposed change will be in place after submitting a new patch for review.

pkg/siteconfig/clusterinstance.go
```diff
@@ -76,8 +143,119 @@ func PullClusterInstance(apiClient *clients.Settings, name, nsname string) (*Clu
...
+	}
+
+	err := wait.PollUntilContextTimeout(
+		context.TODO(), 3*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+			var err error
+			builder.Object, err = builder.Get()
+
+			if err != nil {
+				glog.V(100).Info("failed to get clusterinstance %s/%s: %w",
+					builder.Definition.Name, builder.Definition.Namespace, err)
```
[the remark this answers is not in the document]
mpmaruthu: Okay, updated. Thanks! The proposed change will be in place after submitting a new patch for review.

pkg/siteconfig/clusterinstance.go
```diff
@@ -76,8 +143,119 @@ func PullClusterInstance(apiClient *clients.Settings, name, nsname string) (*Clu
 	return builder, nil
 }
 
+// WithExtraManifests includes manifests via configmap name.
+func (builder *CIBuilder) WithExtraManifests(extraManifestsName string) *CIBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return nil
+	}
+
```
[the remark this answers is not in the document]
mpmaruthu: Okay, updated. Thanks! The proposed change will be in place after submitting a new patch for review.

pkg/siteconfig/clusterinstance.go
```diff
@@ -26,8 +34,67 @@ type ClusterInstanceBuilder struct {
 	errorMsg string
 }
 
+// NewCIBuilder creates a new instance of CIBuilder.
+func NewCIBuilder(apiClient *clients.Settings, name, nsname string, labels map[string]map[string]string) *CIBuilder {
```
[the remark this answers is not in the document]
mpmaruthu: Okay, updated. Thanks!

pkg/siteconfig/clusterinstance_test.go
```diff
@@ -21,6 +28,62 @@ var testSchemes = []clients.SchemeAttacher{
 	siteconfigv1alpha1.AddToScheme,
 }
 
+func TestNewClusterInstanceBuilder(t *testing.T) {
+	generateCIBuilder := NewCIBuilder
```
klaskosk, line 32: You can just use `NewCIBuilder` directly in the test instead of creating the `generateCIBuilder` variable

pkg/siteconfig/clusterinstance_test.go
```diff
@@ -21,6 +28,62 @@ var testSchemes = []clients.SchemeAttacher{
...
+		var testSettings *clients.Settings
+
+		if testCase.client {
+			testSettings = clients.GetTestClients(clients.TestClientParams{})
+		}
+
+		testClusterInstanceStructure := generateCIBuilder(
+			testSettings,
+			testClusterInstance,
+			testClusterInstance)
```
klaskosk, line 76: these should be `testCase.name` and `testCase.namespace`, respectively, which should fix the test failure

pkg/siteconfig/clusterinstance.go
```diff
@@ -76,8 +129,121 @@ func PullClusterInstance(apiClient *clients.Settings, name, nsname string) (*Clu
 	return builder, nil
 }
 
+// WithExtraManifests includes manifests via configmap name.
+func (builder *CIBuilder) WithExtraManifests(extraManifestsName string) *CIBuilder {
+	if valid, _ := builder.validate(); !valid {
+		return nil
```
klaskosk, line 135: apologies for not noticing earlier, this should be `return builder`, which should fix the panic in the unit tests

klaskosk:
lgtm, once the linter error is fixed I'll approve

klaskosk (approved):
/lgtm

PR #706, opened 2024-09-30
pdb: add poddisruptionbudget pkg

sebrandon1:
Adds `poddisruptionbudget` as a package in eco-goinfra.

### DIFF
--- a/pkg/poddisruptionbudget/poddisruptionbudget.go
+++ b/pkg/poddisruptionbudget/poddisruptionbudget.go
@@ -10,6 +10,7 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	schema "k8s.io/apimachinery/pkg/runtime/schema"
 	policyv1typed "k8s.io/client-go/kubernetes/typed/policy/v1"
 )
 
@@ -223,6 +224,15 @@ func (builder *Builder) Update(force bool) (*Builder, error) {
 	builder.Object = builder.Definition
 
 	return builder, err
+}
+
+// GetGVR returns the GroupVersionResource for the Pod Disruption Budget.
+func GetGVR() schema.GroupVersionResource {
+	return schema.GroupVersionResource{
+		Group:    "policy",
+		Version:  "v1",
+		Resource: "poddisruptionbudgets",
+	}
 }
 
 func (builder *Builder) validate() (bool, error) {
--- a/pkg/poddisruptionbudget/poddisruptionbudget.go
+++ b/pkg/poddisruptionbudget/poddisruptionbudget.go
@@ -1,0 +1,254 @@
+package poddisruptionbudget
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	policyv1 "k8s.io/api/policy/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	policyv1typed "k8s.io/client-go/kubernetes/typed/policy/v1"
+)
+
+// Builder provides a struct for the PodDisruptionBudget object and definition.
+type Builder struct {
+	// PodDisruptionBudget definition
+	Definition *policyv1.PodDisruptionBudget
+
+	// Created PodDisruptionBudget object
+	Object *policyv1.PodDisruptionBudget
+
+	// Used in functions that define or mutate PodDisruptionBudget definition.
+	// errorMsg is processed before the PodDisruptionBudget
+	errorMsg  string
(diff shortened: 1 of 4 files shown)


pkg/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return err == nil || !k8serrors.IsNotFound(err)
+}
+
+// WithPDBSpec sets the PodDisruptionBudgetSpec for the PodDisruptionBudget.
+func (builder *Builder) WithPDBSpec(spec policyv1.PodDisruptionBudgetSpec) *Builder {
+	if valid, _ := builder.validate(); !valid {
+		return builder
+	}
+
```
klaskosk, line 179: add a glog here please

pkg/poddisruptionbudget/poddisruptionbudget.go
```diff
@@ -0,0 +1,254 @@
...
+			err := builder.Delete()
+			if err != nil {
+				glog.V(100).Infof(msg.FailToUpdateError("pod disruption budget",
+					builder.Definition.Name, builder.Definition.Namespace))
+			}
+
+			return nil, err
+		}
+
+		return builder.Create()
```
klaskosk, line 217: This return needs to be swapped with the one on line 214

pkg/poddisruptionbudget/poddisruptionbudget_test.go
```diff
@@ -0,0 +1,374 @@
...
+			}
+		} else {
+			assert.Nil(t, err)
+			assert.Equal(t, testResourceQuota.Name, builderResult.Object.Name)
+			assert.Equal(t, testResourceQuota.Namespace, builderResult.Object.Namespace)
+		}
+	}
+}
+
+func TestResourceQuotaWithQuotaSpec(t *testing.T) {
```
klaskosk, line 166: function needs renamed for this resource

pkg/poddisruptionbudget/poddisruptionbudget_test.go
```diff
@@ -0,0 +1,374 @@
...
+		}
+
+		testBuilder := buildTestBuilderWithFakeObjects(runtimeObjects, testPDB.Name, testPDB.Namespace)
+		err := testBuilder.Delete()
+
+		assert.Nil(t, err)
+	}
+}
+
+func TestResourceQuotaExists(t *testing.T) {
```
klaskosk, line 287: function needs renamed for this resource

pkg/poddisruptionbudget/poddisruptionbudget_test.go
```diff
@@ -0,0 +1,374 @@
...
+		assert.Equal(t, testCase.pdbExistsAlready, result)
+	}
+}
+
+func TestPDBUpdate(t *testing.T) {
+	testCases := []struct {
+		rqExistsAlready bool
+		force           bool
+	}{
+		// {
```
klaskosk, line 320: Could you uncomment these please?

pkg/poddisruptionbudget/poddisruptionbudget.go
sebrandon1: Done.

pkg/poddisruptionbudget/poddisruptionbudget.go
sebrandon1: Done.

pkg/poddisruptionbudget/poddisruptionbudget_test.go
sebrandon1: Done.

pkg/poddisruptionbudget/poddisruptionbudget_test.go
sebrandon1: Done.

pkg/poddisruptionbudget/poddisruptionbudget.go
klaskosk: @sebrandon1 I think these still need to be swapped so the `return builder.Create()` is inside the `if force`

sebrandon1: Okay I moved this and also the `resourcequotas.go` was wrong as well.

pkg/resourcequotas/resourcequotas.go
```diff
@@ -153,10 +153,10 @@ func (builder *Builder) Update(force bool) (*Builder, error) {
 				glog.V(100).Infof(msg.FailToUpdateError("resource quota",
 					builder.Definition.Name, builder.Definition.Namespace))
 
-				return nil, err
+				return builder.Create()
 			}
 
-			return builder.Create()
+			return nil, err
```
klaskosk, line 159: @sebrandon1  I think this was fine before, I usually add an extra `return nil, err` outside of the `if force` but inside the `if err != nil` like [here in kac.go](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/ocm/kac.go#L222-L228).

The issue with this change is it puts the `builder.Create()` inside the inner `if err != nil` which means it would only recreate the resource if the delete also failed, not when the delete is successful

klaskosk (approved):
/lgtm

kononovn (approved):
/lgtm

PR #726, opened 2024-10-15
cgu: adjust delete() returns

sebrandon1:
Broken out from: #379 

Making these into smaller PRs for easy review-ability.

### DIFF
--- a/pkg/cgu/cgu.go
+++ b/pkg/cgu/cgu.go
@@ -273,7 +273,12 @@ func (builder *CguBuilder) Delete() (*CguBuilder, error) {
 		builder.Definition.Name, builder.Definition.Namespace)
 
 	if !builder.Exists() {
-		return builder, fmt.Errorf("cgu cannot be deleted because it does not exist")
+		glog.V(100).Infof("cgu %s in namespace %s does not exist",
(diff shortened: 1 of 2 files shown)


pkg/cgu/cgu.go
```diff
@@ -273,7 +273,12 @@ func (builder *CguBuilder) Delete() (*CguBuilder, error) {
 		builder.Definition.Name, builder.Definition.Namespace)
 
 	if !builder.Exists() {
-		return builder, fmt.Errorf("cgu cannot be deleted because it does not exist")
+		glog.V(100).Infof("cgu %s in namespace %s does not exist",
+			builder.Definition.Name, builder.Definition.Namespace)
+
+		builder = nil
```
klaskosk, line 279: I think this should be `builder.Object = nil` instead, the DeleteAndWait test can be changed to have a nil error instead of `fmt.Errorf("error: received nil cgu builder")`

sebrandon1: Done!

klaskosk (approved):
/lgtm

kononovn (approved):
/lgtm

PR #734, opened 2024-10-16
clusterversion: add unit tests and runtime client

klaskosk:
* Added unit tests to the clusterversion package
* Switched the clusterversion package to use the runtime client
* Switched to use constants from configv1 instead of string literals


kononovn (approved):
/lgtm

PR #744, opened 2024-10-21
ci: increase timeout on WaitUntilDeleted()

sebrandon1:
`time.Second` is too short and is causing flapping in the CI test runs.

### DIFF
--- a/pkg/cgu/cgu_test.go
+++ b/pkg/cgu/cgu_test.go
@@ -434,7 +434,7 @@ func TestCguWaitUntilDeleted(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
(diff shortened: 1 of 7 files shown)


klaskosk:
@sebrandon1 I don't think this needs to be updated universally, just for pod. The issue is the same as for #420, the pod `WaitUntilDeleted` uses a `PollUntilContextTimeout` with the immediate parameter set to false which causes the occasional failures. Most other implementations of `WaitUntilDeleted` set immediate to true which prevents the issue

sebrandon1:
Sounds good!  I adjusted the PR to only be `2 * time.Second` and only for the `pod` pkg.

klaskosk (approved):
/lgtm

PR #750, opened 2024-10-21
fmt: fix error return(s)

sebrandon1:
Seems to be an extraneous `fmt.Errorf` on these two packages `daemonset` and `replicaset`.


klaskosk (approved):
/lgtm

PR #727, opened 2024-10-15
clusterlogforwarder: adjust delete() returns

sebrandon1:
Broken out from: #379 

Making these into smaller PRs for easy review-ability.

PR #755, opened 2024-10-21
clusterlogging: add errorMsg check to validate

klaskosk:
The validate method of the clusterLogForwarderBuilder currently does not check for the errorMsg of the builder being set. This PR adds a check for the errorMsg and update the corresponding tests.

Following up on #727.


kononovn (approved):
/lgtm

PR #761, opened 2024-10-22
clients: add extra log on error and use glog

klaskosk:
It turns out there is a case in the clients package where the New function will return nil due to an error but not log the error anywhere. This PR adds an additional log for this code path as well as switches the log statements to use glog instead of the standard library's log package. glog is used elsewhere in the package so this makes it consistent with the package and the rest of the repo.

### DIFF
--- a/pkg/clients/clients.go
+++ b/pkg/clients/clients.go
@@ -2,7 +2,6 @@ package clients
 
 import (
 	"fmt"
-	"log"
 	"os"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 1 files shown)


kononovn:
@klaskosk Did you test it? Do you see log printed in console?

trewest (approved):
/lgtm

klaskosk:
@kononovn I just tested this with eco-gotests and it'll show the log message (provided `ECO_VERBOSE_LEVEL=100` is set)

PR #763, opened 2024-10-25
ibgu: add additional log in Get and namespaces to logs

klaskosk:
* Update all logs to include the namespace of the IBGU since it is a namespaced resource
* Adds an additional log statement for when the Get function fails

Without the additional log when Get fails, it is impossible to tell if the Exist function is returning true because the object truly does not exist or if it returns true because the get failed for a reason other than not found.

An example of this is if the apiClient does not have the CRD for the ibgu, Exist can return true even though that is clearly not the case, causing weird behaviors elsewhere. This PR at least makes the true error more visible in logs.


kononovn (approved):
/lgtm

PR #765, opened 2024-10-31
network: add unit tests and switch to go client

klaskosk:
* Add unit tests for network.go in the network package.
* Switch network.go to use the runtime client and add the Get method.


kononovn (approved):
/lgtm

PR #766, opened 2024-10-31
network: operator unit tests

klaskosk:
Depends-on: openshift-kni/eco-goinfra#765

* Added unit tests to operator.go in the network package, except for the Set* functions since they are effectively untestable in our current setup.
* Made sure scheme was being added in the Pull function
* A few naming edits

### DIFF
--- a/pkg/network/operator_test.go
+++ b/pkg/network/operator_test.go
@@ -1,0 +1,299 @@
+package network
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/network/operator_test.go
```diff
@@ -0,0 +1,299 @@
...
+			K8sMockObjects:  runtimeObjects,
+			SchemeAttachers: operatorTestSchemes,
+		})
+		testBuilder := newOperatorBuilder(testSettings)
+
+		err := testBuilder.WaitUntilInCondition(operatorv1.OperatorStatusTypeAvailable, time.Second, operatorv1.ConditionTrue)
+		assert.Equal(t, testCase.expectedError, err)
+	}
+}
+
```
kononovn, line 197: Please add test cases for:
SetLocalGWMode
SetMultiNetworkPolicy

kononovn (changes requested):
Please add 2 test functions

pkg/network/operator_test.go
klaskosk: @kononovn I'm honestly not sure the best approach to testing these since we really can't cover the functions without modifying the conditions while the function is running. Should I start a goroutine to update the conditions?

kononovn: ok just ignore my comment.

kononovn (approved):
/lgtm

PR #767, opened 2024-10-31
pod: fix ci failure from too short time out

klaskosk:
Similar to #744 except it covers the DeleteAndWait test, which internally uses WaitUntilDeleted and therefore needs to have its timeout extended to 2 seconds for the same reason.


sebrandon1 (approved):

kononovn (approved):
/lgtm

kononovn:
@klaskosk Please re-base

klaskosk:
I didn't even realize but @sebrandon1 already addressed this in #740, so I'll close this

PR #772, opened 2024-11-07
Add managedcluster Create and WithHubAcceptsClient methods

trewest:
- Add Create and WithHubAcceptsClient methods to managedcluster struct
- Add unit tests for Create and WithHubAcceptsClient

### DIFF
--- a/pkg/ocm/managedcluster_test.go
+++ b/pkg/ocm/managedcluster_test.go
@@ -172,6 +172,32 @@ func TestPullManagedCluster(t *testing.T) {
 	}
 }
 
+func TestWithHubAcceptsClient(t *testing.T) {
+	testCases := []struct {
+		accept            bool
+		expectedErrorText string
+	}{
+		{
+			accept:            true,
+			expectedErrorText: "",
+		},
+		{
+			accept:            false,
+			expectedErrorText: "",
+		},
+	}
+
+	for _, testCase := range testCases {
+		testSettings := buildTestClientWithManagedClusterScheme()
+		managedClusterBuilder := buildValidManagedClusterTestBuilder(testSettings).WithHubAcceptsClient(testCase.accept)
+		assert.Equal(t, testCase.expectedErrorText, managedClusterBuilder.errorMsg)
+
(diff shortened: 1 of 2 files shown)


trewest:
@sebrandon1 @klaskosk

pkg/ocm/managedcluster_test.go
```diff
@@ -172,6 +172,32 @@ func TestPullManagedCluster(t *testing.T) {
 	}
 }
 
+func TestWithHubAcceptsClient(t *testing.T) {
+	testCases := []struct {
+		accept            bool
+		expectedErrorText string
```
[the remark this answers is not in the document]
trewest: Not really, I never expect an error message to be returned from the method so I put it explicitly in the test. Otherwise if a change were made that modified the error in some way this test would never see it.

pkg/ocm/managedcluster_test.go
```diff
@@ -172,6 +172,32 @@ func TestPullManagedCluster(t *testing.T) {
...
+		},
+	}
+
+	for _, testCase := range testCases {
+		testSettings := buildTestClientWithManagedClusterScheme()
+		managedClusterBuilder := buildValidManagedClusterTestBuilder(testSettings).WithHubAcceptsClient(testCase.accept)
+		assert.Equal(t, testCase.expectedErrorText, managedClusterBuilder.errorMsg)
+
+		if testCase.expectedErrorText == "" {
+			assert.Equal(t, managedClusterBuilder.Definition.Spec.HubAcceptsClient, testCase.accept)
```
klaskosk, line 196:
```suggestion
			assert.Equal(t, testCase.accept, managedClusterBuilder.Definition.Spec.HubAcceptsClient)
```

klaskosk (commented):
minor comment since the expected comes first in assert.Equal

achuzhoy (dismissed):

pkg/ocm/managedcluster_test.go
trewest: Thanks! Updated

PR #783, opened 2024-11-18
imagestream pkg created to handle imagestream objects

elenagerman:


pkg/clients/clients.go
```diff
@@ -168,6 +169,10 @@ func SetScheme(crScheme *runtime.Scheme) error {
 		return err
 	}
 
+	if err := imagev1.AddToScheme(crScheme); err != nil {
```
klaskosk, line 173: since you're adding the scheme in the Pull function you should be fine without adding it here. You may need to add the scheme in newBuilder too though

pkg/imagestream/imagestream.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the imageStream is empty")
+
+		return nil, fmt.Errorf("imageStream 'nsname' cannot be empty")
+	}
+
+	glog.V(100).Infof(
+		"Pulling imageStream object name %s from namespace %s", name, nsname)
```
klaskosk, line 57: Could you move the log to the top of the function please?

pkg/imagestream/imagestream.go
```diff
@@ -0,0 +1,190 @@
...
+	builder := Builder{
+		apiClient: apiClient.Client,
+		Definition: &imagev1.ImageStream{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: nsname,
+			},
+			Spec: imagev1.ImageStreamSpec{},
+		},
+	}
```
klaskosk, line 68: and it's more just for consistency but could you please move the builder definition to before checking name and nsname parameters?

pkg/imagestream/imagestream.go
```diff
@@ -0,0 +1,190 @@
...
+	if !builder.Exists() {
+		return "", fmt.Errorf("imageStream object does not exist")
+	}
+
+	if len(builder.Object.Spec.Tags) == 0 {
+		return "", fmt.Errorf("imageStream object %s in namespace %s has no tags",
+			builder.Definition.Name, builder.Definition.Namespace)
+	}
+
+	return builder.Object.Spec.Tags[0].From.Name, nil
```
klaskosk, line 137: It looks like From is a pointer and optional so it should have a check that it's not nil so that this function doesn't panic

pkg/imagestream/imagestream_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+		if testCase.client {
+			testSettings = clients.GetTestClients(clients.TestClientParams{
+				K8sMockObjects:  runtimeObjects,
+				SchemeAttachers: testSchemes,
+			})
+		}
+
+		builderResult, err := Pull(testSettings, testCase.name, testCase.namespace)
+		// assert.Equal(t, testCase.expectedError, err)
```
klaskosk, line 101: you should be able to uncomment this and then simplify the if statement below a bit to not check for errors

pkg/imagestream/imagestream_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	for _, testCase := range testCases {
+		exist := testCase.testImageStream.Exists()
+		assert.Equal(t, testCase.expectedStatus, exist)
+	}
+}
+
+func TestImageStreamGetDockerImage(t *testing.T) {
+	testCases := []struct {
```
klaskosk, line 171: could you please add a corresponding test case for checking that Tags[0].From isn't nil?

pkg/imagestream/imagestream_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	for _, testCase := range testCases {
+		dockerImage, err := testCase.testImageStreamBuilder.GetDockerImage()
+		assert.Equal(t, testCase.expectedError, err)
+
+		if testCase.expectedError == nil {
+			assert.Nil(t, err)
+			assert.NotEqual(t, "", dockerImage)
+		} else {
+			assert.Equal(t, testCase.expectedError, err)
```
klaskosk, line 198: since you're checking for the error already, could you please remove the checks from this if statement?

pkg/imagestream/imagestream_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	for _, testCase := range testCases {
+		imageTag, err := testCase.testImageStreamBuilder.GetImageTag()
+		assert.Equal(t, testCase.expectedError, err)
+
+		if testCase.expectedError == nil {
+			assert.Nil(t, err)
+			assert.NotEqual(t, "", imageTag)
+		} else {
+			assert.Equal(t, testCase.expectedError, err)
```
klaskosk, line 231: same as above, could you please remove the error assertions here?

pkg/clients/clients.go
elenagerman: I remove it from the Pull function, it doesn't work vice versa

pkg/imagestream/imagestream.go
klaskosk: @elenagerman Could you still add the validation and leave in the len(Tags) == 0 validation? Even if it's not optional usually, the kubernetes api won't reject it so it's still possible to panic in edge cases which is undesirable

pkg/imagestream/imagestream_test.go
elenagerman: it cannot be nil

elenagerman: if imagestream presented (exists) it's meaning that at least one Tags element is exists

pkg/imagestream/imagestream.go
elenagerman: yes, sure, done

pkg/clients/clients.go
klaskosk: @elenagerman what was the issue having it in the Pull function? we shouldn't need to put it in the clients package anymore

pkg/imagestream/imagestream_test.go
elenagerman: @klaskosk PTAL on the last version

pkg/imagestream/imagestream_test.go
```diff
@@ -0,0 +1,423 @@
...
+			SchemeAttachers: testSchemes,
+		})
+
+		testImageStreamBuilder := buildValidImageStreamBuilder(dummyImageStream)
+
+		dockerImage, err := testImageStreamBuilder.GetDockerImage(testCase.testImageTag)
+		assert.Equal(t, testCase.expectedError, err)
+
+		if testCase.expectedError == nil {
+			assert.Nil(t, err)
```
klaskosk, line 323: since you're already checking that the error and its expected value are equal you can leave out this assertion

pkg/imagestream/imagestream_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+//nolint:funlen
+func TestImageStreamGetDockerImage(t *testing.T) {
+	testCases := []struct {
+		testImageStreamClientWithDummyObject *imagev1.ImageStream
+		testImageTag                         string
+		expectedError                        error
+	}{
+		{
+			testImageStreamClientWithDummyObject: &imagev1.ImageStream{
```
klaskosk, line 177: since a lot of this gets repeated, could you please create a function like `buildDummyImageStreamWithTags(tags []imagev1.TagReference) *imagev1.ImageStream`?

klaskosk (commented):
Thanks for addressing the comments! I've added a couple more minor comments, and then the schemes should be attached in Pull instead of in clients

pkg/clients/clients.go
elenagerman: in general it just fails to create the object: "no kind is registered for the type v1.ImageStream in scheme \"pkg/clients/clients.go:312\"

pkg/imagestream/imagestream_test.go
elenagerman: yes, sure, done

pkg/clients/clients.go
trewest: As @klaskosk mentioned, this should be done in the Pull function. Please reference this example - https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/hive/clusterdeployment.go#L232

trewest: @elenagerman You should be able to do this in the pull function similar to the way it was done here - https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/clusterversion/clusterversion.go#L45. Both of these structs are coming from openshift/api so I imagine they should behave the say way

pkg/imagestream/imagestream.go
```diff
@@ -0,0 +1,180 @@
...
+
+	return err == nil || !k8serrors.IsNotFound(err)
+}
+
+// GetDockerImage fetches imageStream DockerImage value.
+func (builder *Builder) GetDockerImage(imageTag string) (string, error) {
+	if valid, err := builder.validate(); !valid {
+		return "", err
+	}
+
```
trewest, line 118: Can you verify that imageTag is not empty

trewest (changes requested):
Looks good just a couple of comments

pkg/imagestream/imagestream.go
elenagerman: good catch, thank you! missed it; it fixed now

pkg/clients/clients.go
elenagerman: all done: scheme was removed from the client. thanks a lot for advice and provided example!!!

pkg/imagestream/imagestream.go
```diff
@@ -0,0 +1,186 @@
...
+func Pull(apiClient *clients.Settings, name, nsname string) (*Builder, error) {
+	glog.V(100).Infof(
+		"Pulling imageStream object name %s from namespace %s", name, nsname)
+
+	if apiClient == nil {
+		glog.V(100).Infof("The apiClient is empty")
+
+		return nil, fmt.Errorf("imageStream 'apiClient' cannot be empty")
+	}
+
```
klaskosk, line 39: could you please add an AttachScheme here like in [mcp/mcp.go](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/mco/mcp.go#L86-L91)? Unfortunately this won't get caught by our tests but it'll cause errors when you go to use the function if you don't include it

elenagerman: done

klaskosk (approved):
/lgtm

kononovn (approved):
/lgtm

PR #784, opened 2024-11-18
Add log message to clients New function

trewest:
- Add a log message to clients New function to determine why error occurred when loading kubeconfig


trewest:
@klaskosk

PR #793, opened 2024-11-22
ocm: add logs to get errors

klaskosk:
Similar to a change from #763, Exist can return true even if Get returns an error, causing Pull to succeed but Definition to be nil. By adding a log to Get, this error is no longer silent and has a more obvious root cause.


trewest (approved):
/lgtm

PR #798, opened 2024-11-26
velero: add more backup unit tests

klaskosk:
* Added unit tests to cover Get, Exists, Create, Update, and Delete for the Backup builder.
* Updated the returns to be more in line with #355.


kononovn (approved):
/lgtm

PR #802, opened 2024-11-27
assisted agent: adjust delete() returns

sebrandon1:
Broken out from: https://github.com/openshift-kni/eco-goinfra/pull/379


klaskosk (approved):
/lgtm

PR #808, opened 2024-12-05
pod: change wait for pods healthy behavior

klaskosk:
The WaitForAllPodsInNamespacesHealthy function replaces functions in eco-gotests but behaves slightly differently, causing CI instability for cnf/ran.

Previously, the pods were listed every loop then all were checked to see if they were healthy. This means pods that get deleted during the waiting were ignored. Now, however, they cause a failure.

This PR addresses the issue by skipping pods that no longer exist during the waiting period. After all, a pod that does not exist cannot be unhealthy.

### DIFF
--- a/pkg/pod/list.go
+++ b/pkg/pod/list.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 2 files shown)


klaskosk:
cc @yliu127 tested this in my env, should shave 40 minutes off of CI

klaskosk:
cc @mcornea @edcdavid This shouldn't affect anything for ST disk encryption tests since the wait for recover gets retried (whereas for FT it fails eagerly), but let me know if you think this could cause an issue

kononovn (approved):
/lgtm

yliu127:
I'm good with this patch to address the obvious automation issue. 

However, the approach that lists pods only once before entering the loop is not very reliable in my view. There is a chance that new pods will show up after the initial list, and this function can provide false positive results.

yliu127 (approved):

PR #809, opened 2024-12-05
lca ibi: Added WithCABundle function to siteconfig.

achuzhoy:
Added the ability to set clusterinstance.spec.caBundleRef field.


klaskosk (approved):
/lgtm

PR #818, opened 2024-12-16
clusteroperator: adjust return consistency

sebrandon1:


klaskosk (approved):
/lgtm

PR #819, opened 2024-12-16
console: adjust return consistency

sebrandon1:

### DIFF
--- a/pkg/console/console.go
+++ b/pkg/console/console.go
@@ -29,7 +29,7 @@ type Builder struct {
 func NewBuilder(apiClient *clients.Settings, name string) *Builder {
 	glog.V(100).Info("Initializing new console %s structure", name)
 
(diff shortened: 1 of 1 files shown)


pkg/console/console.go
```diff
@@ -62,6 +64,8 @@ func Pull(apiClient *clients.Settings, name string) (*Builder, error) {
 		glog.V(100).Info("The name of the Console is empty")
 
 		builder.errorMsg = "console 'name' cannot be empty"
+
+		return builder, fmt.Errorf("error: %s", builder.errorMsg)
```
klaskosk, line 68:
```suggestion
		return nil, fmt.Errorf("console 'name' cannot be empty")
```
since this is just in Pull I don't think we need builder.errorMsg

klaskosk (approved):
/lgtm

PR #820, opened 2024-12-16
deployment: adjust return consistency

sebrandon1:

No actual returns added here, mostly just changing the references to the pointer.  I find it more consistent to declare variables as a pointer initially then to return a pointer to an object at the end of a function.


klaskosk (approved):
/lgtm

PR #839, opened 2024-12-20
oran: add new oran package and provisioning request builder

klaskosk:
This PR adds the oran package and an initial builder for the ProvisioningRequest resource, along with unit tests.

It makes a number of updates to dependencies:
* Requires oran-o2ims provisioning api
* Indirectly requires the validation dependencies of the provisioning api, which makes up most of the vendoring
* Moves uuid from an indirect to a direct dependency (and allows it in the golangci-lint config)


sebrandon1 (approved):

pkg/oran/provisioningrequest.go
```diff
@@ -0,0 +1,275 @@
...
+		"Initializing new ProvisioningRequest structure with the following params: "+
+			"name: %s, templateName: %s, templateVersion: %s",
+		name, templateName, templateVersion)
+
+	if apiClient == nil {
+		glog.V(100).Infof("The apiClient of the ProvisioningRequest is nil")
+
+		return nil
+	}
+
```
trewest, line 42: Should you check if name, templateName, and templateVersion are empty?

trewest: NVM didn't look down far enough :)

trewest (approved):
/lgtm

PR #844, opened 2025-01-02
namespace: adjust return consistency

sebrandon1:
Broken out from #355


pkg/namespace/namespace.go
```diff
@@ -371,12 +373,14 @@ func (builder *Builder) validate() (bool, error) {
 		glog.V(100).Infof("The %s is undefined", resourceCRD)
 
 		builder.errorMsg = msg.UndefinedCrdObjectErrString(resourceCRD)
```
klaskosk, line 375: now that it returns an error directly could you please remove where it sets errorMsg?

sebrandon1: Done!

klaskosk (approved):
/lgtm

PR #846, opened 2025-01-02
networkpolicy: adjust return consistency

sebrandon1:
Broken out from #355


pkg/networkpolicy/multinetegressrule.go
```diff
@@ -126,10 +126,6 @@ func (builder *EgressRuleBuilder) WithOptions(options ...EgressAdditionalOptions
 func (builder *EgressRuleBuilder) WithPeerPodSelector(podSelector metav1.LabelSelector) *EgressRuleBuilder {
 	glog.V(100).Infof("Adding peer pod selector %v to EgressRule", podSelector)
 
-	if builder.errorMsg != "" {
```
klaskosk, line 129: I think we should be using validate instead of checking the errorMsg here and then we don't have to change the unit tests

sebrandon1: I added a bunch of `validate` calls to the With... funcs.

klaskosk (approved):
/lgtm

PR #847, opened 2025-01-02
nfd: adjust return consistency

sebrandon1:
Broken out from #355 

I had to re-arrange the NewBuilder func to properly handle errors which then forced an adjustment of the unit test cases.

### DIFF
--- a/pkg/nfd/nodefeaturediscovery.go
+++ b/pkg/nfd/nodefeaturediscovery.go
@@ -55,7 +55,7 @@ func NewBuilderFromObjectString(apiClient *clients.Settings, almExample string) 
 		glog.V(100).Infof(
 			"Error initializing NodeFeatureDiscovery from alm-examples: %s", err.Error())
 
-		builder.errorMsg = fmt.Sprintf("Error initializing NodeFeatureDiscovery from alm-examples: %s",
+		builder.errorMsg = fmt.Sprintf("error initializing NodeFeatureDiscovery from alm-examples: %s",
 			err.Error())
 
 		return builder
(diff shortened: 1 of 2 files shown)


pkg/nfd/nodefeaturediscovery.go
```diff
@@ -46,31 +46,35 @@ func NewBuilderFromObjectString(apiClient *clients.Settings, almExample string)
...
+	builder := &Builder{
+		apiClient: apiClient,
 	}
 
+	nodeFeatureDiscovery, err := getNodeFeatureDiscoveryFromAlmExample(almExample)
 	if err != nil {
 		glog.V(100).Infof(
 			"Error initializing NodeFeatureDiscovery from alm-examples: %s", err.Error())
 
 		builder.errorMsg = fmt.Sprintf("Error initializing NodeFeatureDiscovery from alm-examples: %s",
```
klaskosk, line 58:
```suggestion
		builder.errorMsg = fmt.Sprintf("error initializing NodeFeatureDiscovery from alm-examples: %s",
```

pkg/nfd/nodefeaturediscovery.go
```diff
@@ -46,31 +46,35 @@ func NewBuilderFromObjectString(apiClient *clients.Settings, almExample string)
...
 
+	builder.Definition = nodeFeatureDiscovery
+
+	glog.V(100).Infof(
+		"Initializing Builder definition to NodeFeatureDiscovery object")
+
 	if builder.Definition == nil {
 		glog.V(100).Infof("The NodeFeatureDiscovery object definition is nil")
 
 		builder.errorMsg = "NodeFeatureDiscovery definition is nil"
```
klaskosk, line 72:
```suggestion
		builder.errorMsg = "nodeFeatureDiscovery definition is nil"
```

pkg/nfd/nodefeaturediscovery.go
sebrandon1: Done.

pkg/nfd/nodefeaturediscovery.go
sebrandon1: Done.

klaskosk (approved):
/lgtm

PR #850, opened 2025-01-02
nvidiagpu: adjust return consistency

sebrandon1:
Broken out from #355


pkg/nvidiagpu/clusterpolicy.go
```diff
@@ -191,16 +193,15 @@ func (builder *Builder) Create() (*Builder, error) {
...
-		err = builder.apiClient.Create(context.TODO(), builder.Definition)
+		err := builder.apiClient.Create(context.TODO(), builder.Definition)
 
 		if err == nil {
 			builder.Object = builder.Definition
 		}
 	}
 
-	return builder, err
+	return builder, nil
```
klaskosk, line 204: removing the err return here means that the function always returns a nil err. I've tried to write Create to use early returns ([example](https://github.com/klaskosk/eco-goinfra/blob/add-csr/pkg/certificate/signingrequest.go#L107)) to avoid this but it's inconsistent throughout the repo now

sebrandon1: Okay sounds good I can change it back and maybe another PR for the inconsistent Create()'s in the future.

klaskosk (approved):
/lgtm

PR #855, opened 2025-01-03
nmstate: update create() and returns

sebrandon1:
Follow up to: https://github.com/openshift-kni/eco-goinfra/pull/848

Apparently I missed the `Validate()` func when updating the returns.  I also updated the `Create()` to match the example scheme from @klaskosk.


klaskosk (approved):
/lgtm

PR #856, opened 2025-01-03
performanceprofile: adjust return consistency

sebrandon1:
Follow up to: https://github.com/openshift-kni/eco-goinfra/pull/851

The `performanceprofile` file was missed from #355.


klaskosk (approved):
/lgtm

PR #854, opened 2025-01-03
Update GolangCI to v1.63.3

sebrandon1:
Related to: https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2667

PR #860, opened 2025-01-03
rbac: adjust return consistency

sebrandon1:
Broken out from #355


sebrandon1:
This is failing because of #854 needs to merge first.

klaskosk (approved):
/lgtm

PR #873, opened 2025-01-14
Update fec scheme for v2 api

mcornea:
Updating sriovfec.intel.com group version to match the version the operator creates:

```
oc api-resources | grep sriovfec
sriovfecclusterconfigs                sfcc                                                                                   sriovfec.intel.com/v2                         true         SriovFecClusterConfig
sriovfecnodeconfigs                   sfnc                                                                                   sriovfec.intel.com/v2                         true         SriovFecNodeConfig
```


klaskosk:
@mcornea It looks like this code is automatically synced using [this config](https://github.com/openshift-kni/eco-goinfra/blob/main/internal/sync/configs/fec-config.yaml) so we can't change it manually, we'd have to change that config

mcornea:
> @mcornea It looks like this code is automatically synced using [this config](https://github.com/openshift-kni/eco-goinfra/blob/main/internal/sync/configs/fec-config.yaml) so we can't change it manually, we'd have to change that config

Thanks @klaskosk I updated the files using the sync tool.

klaskosk (approved):
/lgtm

sebrandon1 (approved):

PR #878, opened 2025-01-20
hive clusterimageset: return consistency

sebrandon1:
Follow up to: #822 

These were missed as part of this PR.


klaskosk (approved):
/lgtm

PR #880, opened 2025-01-20
validatingwebhook: adjust return consistency

sebrandon1:
Follow up to: https://github.com/openshift-kni/eco-goinfra/pull/869

These were missed in the original PR.


klaskosk (approved):
/lgtm

PR #882, opened 2025-01-21
adding nfd feature rule package and new builder

ggordaniRed:
Adding new lib,builder,unit tests for nodfeaturerule object


cdvultur (changes requested):
some minor changes

go.mod
```diff
@@ -53,9 +53,9 @@ require (
 	gopkg.in/k8snetworkplumbingwg/multus-cni.v4 v4.0.2
 	gopkg.in/yaml.v2 v2.4.0
 	gorm.io/gorm v1.24.5
-	k8s.io/api v0.30.5
+	k8s.io/api v0.30.7
```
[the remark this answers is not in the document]
ggordaniRed: I add new package by run make and make install maybe in the process this lib version changed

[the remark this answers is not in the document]
cdvultur: @sebrandon1 It seems that NFD is not part of the sync job, thus it is causing this problem. What is currently the procedure? Should @ggordaniRed add config for the sync ?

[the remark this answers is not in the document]
klaskosk: We shouldn't need a sync config for this operator unless it's causing issues with dependencies. There doesn't seem to be any actual changes between 0.30.5 and 0.30.7 of these libraries though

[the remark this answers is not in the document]
sebrandon1: Okay sounds good.  👍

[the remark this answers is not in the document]
ggordaniRed: @sebrandon1 @klaskosk So should I revert or can we mere as it is cause resolved all other comments so let me know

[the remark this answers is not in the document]
klaskosk: @ggordaniRed I'll go ahead and review but this version change is fine

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(nodeFeatureRuleList.Items) == 0 {
+		return nil, fmt.Errorf("failed to get alm examples")
+	}
+
+	for i, item := range nodeFeatureRuleList.Items {
+		if item.Kind == "NodeFeatureRule" {
```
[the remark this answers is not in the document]
ggordaniRed: basically there are two kind of items or 3 but this is the same code that run on NodeFeatureDiscovery pkg and I test this and didn't encounter any runtime or marshaling error from the Json parser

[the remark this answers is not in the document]
ggordaniRed: I think because the struct look similar I'll get more than one object I thought using maps instead but it really unnecessary and instead I just make sure that I find the right index this is why I had this if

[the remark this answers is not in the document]
klaskosk: @ggordaniRed ah that's my bad, I misunderstood how the Unmarshal works. As long as there are no conflicting fields of different types we should be fine

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+		nodeFeatureRuleBuilder.errorMsg = "NodeFeatureRule definition is nil"
+
+		return &nodeFeatureRuleBuilder
+	}
+
+	return &nodeFeatureRuleBuilder
+}
+
+// FeatureRulePull loads an existing NodeFeatureRuleBuilder into Builder struct.
+func FeatureRulePull(apiClient *clients.Settings, name, namespace string) (*NodeFeatureRuleBuilder, error) {
```
klaskosk, line 81:
```suggestion
func PullFeatureRule(apiClient *clients.Settings, name, namespace string) (*NodeFeatureRuleBuilder, error) {
```
consistency with other packages

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+	glog.V(100).Infof("Pulling existing NodeFeatureRule name: %s in namespace: %s", name, namespace)
+
+	if apiClient == nil {
+		glog.V(100).Info("The apiClient of the NodeFeatureRule is nil")
+
+		return nil, fmt.Errorf("the apiClient of the Policy is nil")
```
klaskosk, line 87:
```suggestion
		return nil, fmt.Errorf("the apiClient of the NodeFeatureRule is nil")
```

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+				Name:      name,
+				Namespace: namespace,
+			},
+		},
+	}
+
+	if name == "" {
+		glog.V(100).Infof("NodeFeatureRule name is empty")
+
+		return nil, fmt.Errorf("NodeFeatureRule 'name' cannot be empty")
```
klaskosk, line 110:
```suggestion
		return nil, fmt.Errorf("nodeFeatureRule 'name' cannot be empty")
```

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	if namespace == "" {
+		glog.V(100).Infof("NodeFeatureRule namespace is empty")
+
+		return nil, fmt.Errorf("NodeFeatureRule 'namespace' cannot be empty")
```
klaskosk, line 116:
```suggestion
		return nil, fmt.Errorf("nodeFeatureRule 'namespace' cannot be empty")
```

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	if !ruleBuilder.Exists() {
+		return nil, fmt.Errorf("NodeFeatureRule object %s does not exist in namespace %s", name, namespace)
```
klaskosk, line 120:
```suggestion
		return nil, fmt.Errorf("nodeFeatureRule object %s does not exist in namespace %s", name, namespace)
```

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	ruleBuilder.Definition = ruleBuilder.Object
+
+	return &ruleBuilder, nil
```
klaskosk, line 125: see above about returns consistency (#355)
```suggestion
	return ruleBuilder, nil
```

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+
+		return false, fmt.Errorf(msg.UndefinedCrdObjectErrString(resourceCRD))
+	}
+
+	if builder.apiClient == nil {
+		glog.V(100).Infof("The %s builder apiclient is nil", resourceCRD)
+
+		return false, fmt.Errorf("%s builder cannot have nil apiClient", resourceCRD)
+	}
+
```
klaskosk, line 248: please add a check for errorMsg ([example](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/bmh/hfs.go#L189-L193))

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,198 @@
...
+	errorMsg string
+}
+
+// NewnodeFeatureRuleBuilderFromObjectString creates a Builder object from CSV alm-examples.
+func NewnodeFeatureRuleBuilderFromObjectString(apiClient *clients.Settings, almExample string) *NodeFeatureRuleBuilder {
+	glog.V(100).Infof(
+		"Initializing new Builder structure from almExample string")
+
+	if apiClient == nil {
+		glog.V(100).Info("The apiClient of the Policy is nil")
```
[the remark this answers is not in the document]
klaskosk: This still seems like it needs to be changed

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,250 @@
...
+// NodeFeatureRuleBuilder provides a struct for NodeFeatureRule object
+// from the cluster and a NodeFeatureRule definition.
+type NodeFeatureRuleBuilder struct {
+	// Builder definition. Used to create
+	// Builder object with minimum set of required elements.
+	Definition *nfdv1.NodeFeatureRule
+	// Created Builder object on the cluster.
+	Object *nfdv1.NodeFeatureRule
+	// api client to interact with the cluster.
+	apiClient *clients.Settings
```
klaskosk, line 26: another thing I missed, but this should be switched to the client from controller-runtime. We accept `*clients.Settings` in functions but only store `apiClient.Client` in the builder ([example](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/bmh/hfs.go#L22))

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,198 @@
...
+			builder.Definition.Name, builder.Definition.Namespace)
+
+		return nil, err
+	}
+
+	return NodeFeatureRule, err
+}
+
+// GetErrorMessage returns builder's error message.
+func (builder *NodeFeatureRuleBuilder) GetErrorMessage() string {
```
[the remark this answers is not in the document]
klaskosk: could you please address this?

klaskosk (commented):
Thanks for addressing the comments! I have a few more minor comments but overall is looking pretty good

klaskosk:
@ggordaniRed Thanks for making the changes, I unresolved a few conversations that still need to be addressed though. I'm not sure why but I think the changes to address the comments are getting overwritten by a later commit. My guess would be it's related to accepting suggestions and then committing from your local branch but not fully sure

pkg/nfd/nodefeaturerule_test.go
```diff
@@ -0,0 +1,229 @@
+package nfd_test
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+	"testing"
+
+	. "github.com/openshift-kni/eco-goinfra/pkg/nfd"
```
klaskosk, line 9: I hadn't noticed till the linter called it out but you should change the package to `nfd` even for the test then you won't have to dot import anything

pkg/nfd/nodefeaturerule.go
```diff
@@ -0,0 +1,253 @@
...
+
+		return &nodeFeatureRuleBuilder
+	}
+
+	if nodeFeatureRuleBuilder.Definition == nil {
+		glog.V(100).Infof("The NodeFeatureRule object definition is nil")
+
+		nodeFeatureRuleBuilder.errorMsg = "nodeFeatureRule definition is nil"
+		
+		return builder
```
klaskosk, line 74: I think I wrote the suggestion wrong out of force of habit but this should be `nodeFeatureRuleBuilder` and then all other returns should be changed to remove the `&`

ggordaniRed: @klaskosk  I just need to return a ref or pointer so should I put "&" in return statement or in variable initialization?

klaskosk: @ggordaniRed Put it in the variable initialization please

pkg/nfd/nodefeaturerule_test.go
klaskosk: @ggordaniRed sorry for the confusion, but I meant change the package of the file, not the alias you import it as. In eco-goinfra we keep the tests in the same package as the implementations as it allows us to check things like validate() functions and builder.errorMsg

pkg/nfd/nodefeaturerule.go
klaskosk: @ggordaniRed It looks like it's still in the return statement, not the variable initialization

go.mod
```diff
@@ -214,6 +214,8 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 )
 
+require github.com/openshift/node-feature-discovery/api/nfd v0.0.0-20250103121957-354b94ed3e55
```
cdvultur, line 217: Not sure why this is left outside the 2 `require` blocks above . Can you move it to the first one ?

cdvultur (changes requested):
Small comment on the go.mod

ggordaniRed:
Open a more cleaner PR instead

PR #884, opened 2025-01-21
assisted: adjust return consistency

sebrandon1:
Broken out from #355 

Follow up to #815 where these were missed.


klaskosk (approved):
/lgtm

PR #891, opened 2025-01-23
namespace: Update CleanObjects

ajaggapa:
`CleanObjects` uses graceperiod=0 which is not recommended to have smooth deletion of resources. Hence, removing `GracePeriodSeconds` parameter.


klaskosk (approved):
/lgtm
In eco-gotests, only the cnf network team uses this function so there is no risk of this affecting other teams

sebrandon1 (approved):

trewest (approved):
/lgtm

PR #898, opened 2025-01-28
bmc: backport serial console test fix

klaskosk:
The timeout error message in TestBMCSerialConsole takes two different forms. Tests on the main branch already account for this but release-4.16 does not have the fix. This PR replaces the test on release-4.16 with the one from main.


trewest (approved):
/lgtm

PR #904, opened 2025-01-30
poddisruptionbudget pkg: get list option was added

elenagerman:
poddisruptionbudget pkg: get list option was added: for the single namespace and for all namespaces


pkg/poddisruptionbudget/list.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, fmt.Errorf("failed to list podDisruptionBudgets, 'nsname' parameter is empty")
+	}
+
+	logMessage := fmt.Sprintf("Listing podDisruptionBudget in the nsname %s", nsname)
+	passedOptions := metav1.ListOptions{}
+
+	if len(options) >= 1 {
+		passedOptions = options[0]
+		logMessage += fmt.Sprintf(" with the options %v", passedOptions)
+	}
```
yprokule, line 32: if only 1st element is used then better to have explicit check (like in other packages)
```
    if len(options) > 1 {
        glog.V(100).Infof("'options' parameter must be empty or single-valued")

        return nil, fmt.Errorf("error: more than one ListOptions was passed")
    }

    if len(options) == 1 {
        passedOptions = options[0]
        logMessage += fmt.Sprintf(" with the options %v", passedOptions)
    }
```

pkg/poddisruptionbudget/list.go
```diff
@@ -0,0 +1,113 @@
...
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// List returns podDisruptionBudget inventory in the given namespace.
+func List(apiClient *clients.Settings, nsname string, options ...metav1.ListOptions) ([]*Builder, error) {
```
klaskosk, line 13: could you please add a helper to handle the shared listing code like in [bmh/list.go](https://github.com/openshift-kni/eco-goinfra/blob/main/pkg/bmh/list.go#L133)?

elenagerman: sure, thanks!

pkg/poddisruptionbudget/list.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	glog.V(100).Infof(logMessage)
+
+	return list(apiClient, passedOptions)
+}
+
+// list lists the podDisruptionBudget according to the provided options.
+func list(apiClient *clients.Settings, options goclient.ListOptions) ([]*Builder, error) {
+	err := apiClient.AttachScheme(policyv1.AddToScheme)
```
klaskosk, line 78: since the rest of the package uses the PolicyV1Interface instead of goclient, could you use that for list instead? So stick with metav1.ListOptions and then this `list` function could take a namespace parameter that's just empty for ListInAllNamespaces and list like
```go
apiClient.PodDisruptionBudgets(namespace).List(/* ... */)
```

klaskosk (commented):
Sorry the bmh example is the best we have in the repo but not the greatest for this package with the different client

pkg/poddisruptionbudget/list.go
elenagerman: sure, done

sebrandon1 (approved):

PR #905, opened 2025-01-31
bmc: add WaitForSystemPowerState function

klaskosk:
When performing a reset action (power on, power off, reset, etc.), it is useful to verify the action actually took place. WaitForSystemPowerState allows one to poll the BMC until the desired power state occurs, similar to functions already used in cnf/ran in eco-gotests.


kononovn (approved):
/lgtm

PR #909, opened 2025-02-01
pod: remove default GracePeriod

ajaggapa:
[PR](https://github.com/openshift-kni/eco-goinfra/pull/891) ensures grace-period is not set to while terminating pod. But kubernetes uses GracePeriod defined in Pod Spec. However, `TerminationGracePeriodSeconds` in PodSpec is being set to 0 for all new pods defined with `NewBuilder` unless explicitly updated with `WithTerminationGracePeriodSeconds`

Hence, removing grace-period=0 in default pod definition.


klaskosk (approved):
/lgtm

sebrandon1 (approved):

PR #918, opened 2025-02-04
namespace: edit CleanObjects [4.15]

ajaggapa:
CleanObjects is not working when there is only 1 pod/resource in terminating state


klaskosk (approved):
/lgtm

sebrandon1 (approved):

PR #921, opened 2025-02-05
Import consistency

sebrandon1:
Makes sure that imports are using a more descriptive name than `v1`.

Note: This does not touch the `schemes` folder that are copies of other code.

Similar to:
- #239 
- #387 
- #694 
- #711


klaskosk (approved):
/lgtm

PR #928, opened 2025-02-11
pod: add new health functions

klaskosk:
This is part 1 of a 3 part effort to replace pod.WaitForAllPodsInNamespacesHealthy.

1. (This PR) Add new helpers to eco-goinfra that simplify waiting for pods healthy and list on every iteration.
2. Change the cluster.WaitForRecover function to use the new eco-goinfra helpers. Also a small change in cnf/ran/talm.
3. Remove the old helpers from eco-goinfra.

Motivation:

The old WaitForRecover function would list the pods each iteration, ensuring that pods started after calling the function will be checked as well as ignoring pods that later end up deleted. A workaround for the latter of these issues was merged in #808.

Additionally, the current pod.WaitForAllPodsInNamespacesHealthy has too many parameters and is hard to test properly. The current unit tests for it do not adequately cover all scenarios. This function only gets used in one place in eco-gotests, the aforementioned cluster.WaitForRecover, where many parameters will always be the same.

The new eco-goinfra helpers introduced in this PR will notably exclude the ignoreNamespaces option. It is a remnant of an issue specific to the cnf/ran team that no longer exists and can safely be removed to simplify the helpers further.

### DIFF
--- a/pkg/pod/pod_test.go
+++ b/pkg/pod/pod_test.go
@@ -647,6 +647,56 @@ func TestPodWaitUntilInStatuses(t *testing.T) {
 	}
 }
 
+func TestPodIsHealthy(t *testing.T) {
+	testCases := []struct {
+		exists          bool
+		phase           corev1.PodPhase
+		condition       corev1.PodConditionType
+		expectedHealthy bool
+	}{
+		{
+			exists:          true,
+			phase:           corev1.PodSucceeded,
+			condition:       "",
+			expectedHealthy: true,
+		},
+		{
+			exists:          true,
+			phase:           corev1.PodRunning,
+			condition:       corev1.PodReady,
+			expectedHealthy: true,
+		},
+		{
+			exists:          false,
+			phase:           "",
+			condition:       "",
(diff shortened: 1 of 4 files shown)


pkg/pod/pod_test.go
```diff
@@ -647,6 +647,57 @@ func TestPodWaitUntilInStatuses(t *testing.T) {
...
+			runtimeObjects = append(runtimeObjects, testPod)
+		}
+
+		testSettings := clients.GetTestClients(clients.TestClientParams{
+			K8sMockObjects: runtimeObjects,
+		})
+		testBuilder := buildValidPodTestBuilder(testSettings)
+
+		healthy := testBuilder.IsHealthy()
+		assert.Equal(t, testCase.expectedHealthy, healthy)
```
sebrandon1, line 697:
```suggestion
		assert.Equal(t, testCase.expectedHealthy, testBuilder.IsHealthy())
```

sebrandon1 (dismissed):
One minor nit but lgtm!

PR #937, opened 2025-02-20
Update Go to v1.24.1

sebrandon1:
Related to: https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2795

Notable changes: 
- In the actions YAMLs (to make future Go updates more manageable), change the reference to point to the `go.mod` file itself versus having individual references on multiple files.
- Adjusting all of the calls to `fmt.Sprintf` and `fmt.Errorf` to pass the new `go vet`.
- Update golangci-lint to v1.64.2 because 1.63.x does not support Go 1.24.x.
- `tenv` linter was deprecated and replaced with `usetesting`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 run:
-  go: 1.23
+  go: "1.24"
   timeout: 15m0s
   fast: false
 #  modules-download-mode: readonly
@@ -72,7 +72,6 @@ linters-settings:
     statements: 40
   unused:
     check-exported: true
-    go: "1.23"
   staticcheck:
     # https://staticcheck.io/docs/options#checks
     checks:
@@ -124,12 +123,12 @@ linters:
     - revive
     - staticcheck
     - stylecheck
-    - tenv
     - thelper
     - typecheck
     - unconvert
     - unparam
     - unused
+    - usetesting
     - varnamelen
     - wsl
     - nlreturn
--- a/pkg/kmm/preflightvalidationocp.go
+++ b/pkg/kmm/preflightvalidationocp.go
@@ -182,17 +182,13 @@ func PullPreflightValidationOCP(apiClient *clients.Settings,
 	if name == "" {
 		glog.V(100).Infof("The name of the preflightvalidationocp is empty")
 
-		builder.errorMsg = "preflightvalidationocp 'name' cannot be empty"
-
-		return builder, fmt.Errorf(builder.errorMsg)
+		return builder, fmt.Errorf("%s", "preflightvalidationocp 'name' cannot be empty")
 	}
 
 	if nsname == "" {
 		glog.V(100).Infof("The namespace of the preflightvalidationocp is empty")
 
-		builder.errorMsg = "preflightvalidationocp 'nsname' cannot be empty"
-
-		return builder, fmt.Errorf(builder.errorMsg)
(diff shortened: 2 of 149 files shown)


klaskosk:
+1 on switching to use the go mod file for versions

.golangci.yml
```diff
@@ -72,7 +72,7 @@ linters-settings:
     statements: 40
   unused:
     check-exported: true
-    go: "1.23"
+    go: "1.24"
```
klaskosk, line 75: this field isn't an actual configuration option for unused so we can just remove it in the spirit of reducing the number of places we reference versions

.golangci.yml
```diff
@@ -1,5 +1,5 @@
 run:
-  go: 1.23
+  go: 1.24
```
klaskosk, line 2: this should be a string even if it does work as a number

klaskosk (commented):
a couple of minor comments about the lint config I noticed after updating it for eco-gotests

.golangci.yml
klaskosk: or actually the entire section since check-exported isn't either

pkg/storage/pvc.go
```diff
@@ -87,14 +87,14 @@ func (builder *PVCBuilder) WithPVCAccessMode(accessMode string) (*PVCBuilder, er
 		glog.V(100).Infof("Empty accessMode for PVC %s", builder.Definition.Name)
 		builder.errorMsg = "Empty accessMode for PVC requested"
 
-		return builder, fmt.Errorf(builder.errorMsg)
+		return builder, fmt.Errorf("%s", builder.errorMsg)
```
klaskosk, line 90: same as above, could you please switch this to return the error directly?

pkg/storage/pvc.go
```diff
@@ -87,14 +87,14 @@ func (builder *PVCBuilder) WithPVCAccessMode(accessMode string) (*PVCBuilder, er
...
 	}
 
 	if !validatePVCAccessMode(accessMode) {
 		glog.V(100).Infof("Invalid accessMode for PVC %s", accessMode)
 		builder.errorMsg = fmt.Sprintf("Invalid accessMode for PVC %s", accessMode)
 
-		return builder, fmt.Errorf(builder.errorMsg)
+		return builder, fmt.Errorf("%s", builder.errorMsg)
```
klaskosk, line 97: same as above

klaskosk (commented):
a couple minor comments to fix some returns but otherwise lgtm

pkg/kmm/preflightvalidationocp.go
```diff
@@ -184,15 +184,15 @@ func PullPreflightValidationOCP(apiClient *clients.Settings,
 
 		builder.errorMsg = "preflightvalidationocp 'name' cannot be empty"
 
-		return builder, fmt.Errorf(builder.errorMsg)
+		return builder, fmt.Errorf("%s", builder.errorMsg)
```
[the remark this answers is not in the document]
sebrandon1: Good find!  I'll adjust!

klaskosk (approved):
Thanks for making all of these changes! /lgtm

sebrandon1:
Go 1.24.1 is now out, the PR has been updated accordingly: https://github.com/redhat-best-practices-for-k8s/certsuite/pull/2855

klaskosk (approved):

PR #947, opened 2025-03-01
integration: add pod tests

klaskosk:
* Added Create and Delete integration tests for the pod resource.
* Added an ExecCommand test too for the pod resource. Since we do not have any unit tests for this function (and #602 to add them is a draft), it will be good to have some assurance it is working.

The latter point is also motivated by a desire to use a FallbackExecutor to enable the websocket executor for Kubernetes 1.30+ (OpenShift 4.17+). This will be a future PR, however.

### DIFF
--- a/integration/pod_test.go
+++ b/integration/pod_test.go
@@ -1,0 +1,142 @@
+//go:build integration
+// +build integration
+
+package integration
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/namespace"
+	"github.com/openshift-kni/eco-goinfra/pkg/pod"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestPodCreate(t *testing.T) {
+	t.Parallel()
+	client := clients.New("")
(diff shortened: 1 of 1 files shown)


sebrandon1 (dismissed):

integration/pod_test.go
```diff
@@ -0,0 +1,180 @@
...
+	namespaceBuilder := namespace.NewBuilder(client, testNamespace)
+	assert.NotNil(t, namespaceBuilder)
+
+	// Preemptively delete the namespace before the test
+	err := namespaceBuilder.DeleteAndWait(time.Duration(30) * time.Second)
+	assert.Nil(t, err)
+
+	// Create the namespace
+	namespaceBuilder, err = namespaceBuilder.Create()
+	assert.Nil(t, err)
```
sebrandon1, line 88: We can replace this with: https://github.com/openshift-kni/eco-goinfra/blob/main/integration/common.go#L13

integration/pod_test.go
```diff
@@ -0,0 +1,180 @@
...
```
sebrandon1, line 147: Same here, use the `common.go` funcs for the namespace management.  👍

sebrandon1 (approved):

PR #714, opened 2024-10-04
PoC: Add hosted integration testing

sebrandon1:
Work in progress!

PR #949, opened 2025-03-03
integration: add daemonset tests

sebrandon1:
- Added Create and Delete integration tests for the `daemonset` resource.
- I had to modify the `IsDaemonsetReady` to actually match what we do in the `certsuite-qe` repo [here](https://github.com/redhat-best-practices-for-k8s/certsuite-qe/blob/main/tests/globalhelper/daemonset.go#L60-L66).  For some reason, when testing, the `status` wouldn't have conditions listed and it would never come back as ready.  When this merges, I'll change our QE to use the eco-goinfra logic.
- I also added `!k8serrors.IsNotFound(err)` to the daemonset `Delete()` to better handle already "missing" daemonsets.

Related to: 
- #714

### DIFF
--- a/pkg/daemonset/daemonset.go
+++ b/pkg/daemonset/daemonset.go
@@ -321,13 +321,13 @@ func (builder *Builder) Delete() error {
 	err := builder.apiClient.Delete(
 		context.TODO(), builder.Definition.Name, metav1.DeleteOptions{})
 
-	if err != nil {
+	if err != nil && !k8serrors.IsNotFound(err) {
 		return err
 	}
 
 	builder.Object = nil
 
-	return err
+	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/daemonset/daemonset.go
```diff
@@ -321,13 +321,13 @@ func (builder *Builder) Delete() error {
 	err := builder.apiClient.Delete(
 		context.TODO(), builder.Definition.Name, metav1.DeleteOptions{})
 
-	if err != nil {
+	if err != nil && !k8serrors.IsNotFound(err) {
```
klaskosk, line 324: this race condition between our exists check and delete could be present in most delete functions, interesting to keep an eye out for it now that it showed up in tests

klaskosk (approved):

PR #956, opened 2025-03-05
hw-accel: move KMM to use sync for dependecy

cdvultur:
Moving KMM to use the `internal/sync` functionality.
Also bumped it up to use `release-2.3`

Further updates to the builders and functionality will be done in another PR.


klaskosk (approved):
/lgtm

PR #957, opened 2025-03-06
lca: changed returned error in seedgenerator waituntilcomplete

Demostenes777:
The WaitUntilComplete() function was not behaving correctly. It was not pollling until the deadline because the return of the fmt.Errorf("seedgenerator did not complete") was making the loop to stop. This PR changes the returned error when the condition is not met and updates the unit test accordingly.


klaskosk (approved):
/lgtm

kononovn (approved):
/lgtm

PR #958, opened 2025-03-06
integration: add common funcs

sebrandon1:
Adds an `integration/common.go` with two funcs:
- CreateTestContainerDefinition()
- PreEmptiveNamespaceDeleteAndSetup()

Stuff that is "shared" between our various integration tests can be placed here to add more code re-use.

### DIFF
--- a/integration/common.go
+++ b/integration/common.go
@@ -1,0 +1,50 @@
+package integration
+
+import (
+	"errors"
+	"time"
+
+	"github.com/openshift-kni/eco-goinfra/pkg/namespace"
+	"github.com/openshift-kni/eco-goinfra/pkg/pod"
+	corev1 "k8s.io/api/core/v1"
+)
+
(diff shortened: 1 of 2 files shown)


integration/common.go
```diff
@@ -0,0 +1,46 @@
...
+	containerDefinition, err := testContainerBuilder.GetContainerCfg()
+	if err != nil {
+		return nil, err
+	}
+
+	// Change the container default security context to something that is allowed in the test environment
+	testContainerBuilder.WithSecurityContext(&corev1.SecurityContext{
+		RunAsUser:  nil,
+		RunAsGroup: nil,
+	})
```
klaskosk, line 43: I think this makes more sense to put before we `GetContainerCfg` since that way we get the validations run with the security context added

klaskosk (approved):

PR #960, opened 2025-03-06
integration: modify 'deployment' to use common funcs

sebrandon1:
Follow up to #958 

Bonus, I added some `nil` checks to the daemonset tests after creating the container definition.


klaskosk (approved):
/lgtm good catch in the daemonset test

PR #942, opened 2025-02-24
PoC: Add nightly OCP integration test

sebrandon1:
Adds a nightly YAML for triggering both k8s and OCP runs nightly (not tied to any PR).

This runs the exact `make integration-test` path, but against OCP in a nightly fashion.

Brings in changes from:
- #937 
- #961

PR #961, opened 2025-03-06
ci: prevent forks from running nightly crons

sebrandon1:
Resolves: https://github.com/openshift-kni/eco-goinfra/issues/959

If any more `cron` related jobs (eg. #942) get added, they will also need the conditional statement.


trewest (approved):
/lgtm

klaskosk (approved):
/lgtm

PR #964, opened 2025-03-11
WIP: metallb: add unnumbered to bgppeer

gkopels:
Adding unnumbered interface option to the bgppeer spec.


ajaggapa:
eco-gotests integration check will fail with change as there are already tests calling this function with peerip arg.

Instead let the NewBuilder accept peerip argument and have a checkpoint if it is empty or not
if peerip = empty (Unnumbered case) : dont define Address field in BGPPeer 
if not : define Address field in BGPPeer

pkg/metallb/bgppeer.go
```diff
@@ -258,6 +304,60 @@ func (builder *BGPPeerBuilder) Update(force bool) (*BGPPeerBuilder, error) {
...
+		return builder
+	}
+
+	if bgpPeerIP == "" {
+		glog.V(100).Infof("Can not redefine BGPPeer with empty IP address string")
+
+		builder.errorMsg = "ip address can not be empty string"
+
+		return builder
+	}
```
klaskosk, line 331: net.ParseIP will return nil on an empty string so this check is unnecessary

pkg/metallb/bgppeer_test.go
```diff
@@ -52,14 +52,6 @@ func TestNewBPGPeerBuilder(t *testing.T) {
 			remoteAsn:     5002,
 			expectedError: "BGPPeer 'nsname' cannot be empty",
 		},
-		{
```
klaskosk, line 55: could you please switch this test to the new function?

pkg/metallb/bgppeer_test.go
```diff
@@ -504,7 +523,7 @@ func buildValidBGPPeerBuilder(apiClient *clients.Settings) *BGPPeerBuilder {
 }
 
 func buildInValidBGPPeerBuilder(apiClient *clients.Settings) *BGPPeerBuilder {
-	return NewBPGPeerBuilder(apiClient, defaultBGPPeerName, defaultBGPPeerNsName, "", 1000, 2000)
+	return NewBPGPeerBuilder(apiClient, defaultBGPPeerName, defaultBGPPeerNsName, "172.16.20.1000", 1000, 2000)
```
klaskosk, line 526: wouldn't this return a valid builder? could you switch these to the new function as well please

pkg/metallb/bgppeer.go
```diff
@@ -258,6 +304,60 @@ func (builder *BGPPeerBuilder) Update(force bool) (*BGPPeerBuilder, error) {
 	return builder, err
 }
 
+// WithBGPPeerIP defines the peer IP address.
+func (builder *BGPPeerBuilder) WithBGPPeerIP(bgpPeerIP string) *BGPPeerBuilder {
```
klaskosk, line 308: could you please add a unit test for this function?

klaskosk (commented):
a few comments, in general I think it's better to switch the tests to the new function now that way it's tested before eco-gotests is switched over

pkg/metallb/bgppeer.go
gkopels: added

pkg/metallb/bgppeer.go
gkopels: removed

pkg/metallb/bgppeer_test.go
gkopels: switched

pkg/metallb/bgppeer_test.go
gkopels: Switched

pkg/metallb/bgppeer.go
```diff
@@ -84,6 +84,52 @@ func NewBPGPeerBuilder(
 	return builder
 }
 
+// NewBGPPeerBuilder creates a new instance of BGPPeer.
+func NewBGPPeerBuilder(
+	apiClient *clients.Settings, name, nsname string, asn, remoteASN uint32) *BGPPeerBuilder {
+	glog.V(100).Infof(
+		"Initializing new BGPPeer structure with the following params: %s, %s %s %d %d",
+		name, nsname, asn, remoteASN)
+
```
klaskosk, line 93: I missed this the first round, but could you please add a check for nil apiClient?
```go
if apiClient == nil {
        glog.V(100).Info("BGPPeer 'apiClient' cannot be nil")

        return nil
}
```

klaskosk (commented):
one thing I missed otherwise lgtm

pkg/metallb/bgppeer.go
```diff
@@ -84,6 +84,52 @@ func NewBPGPeerBuilder(
...
+		return builder
+	}
+
+	if nsname == "" {
+		glog.V(100).Infof("The namespace of the BGPPeer is empty")
+
+		builder.errorMsg = "BGPPeer 'nsname' cannot be empty"
+
+		return builder
+	}
```
ajaggapa, line 128: Have these checks before updating builder with values (line101)

pkg/metallb/bgppeer.go
gkopels: added

pkg/metallb/bgppeer.go
klaskosk: the convention at least for other packages is to do the checks after creating the builder with name and nsname

klaskosk (dismissed):

sebrandon1 (changes requested):
`TestBGPPeerWithRouterID` should set the `Name` field in the object struct to allow for testing of the other logic.

klaskosk (approved):

PR #972, opened 2025-03-14
bmc: added bootfromcd function

Demostenes777:
This PR adds a function to the bmc package to boot from CD only once. This functionality is required for the ORAN system tests based on the IBI-based provisioning.


klaskosk (approved):
/lgtm

PR #980, opened 2025-03-25
ci: remove script call; update golangci-lint

sebrandon1:
This was mistakenly added in #977 and is not needed.


klaskosk:
I think we're failing due to #979, I need to look into how we the action triggers CI since I don't think it's right

klaskosk (dismissed):

cdvultur (dismissed):
/lgtm

PR #988, opened 2025-03-31
DeleteGraceful method for deployment added

yprokule:


pkg/deployment/deployment.go
```diff
@@ -484,6 +484,45 @@ func (builder *Builder) Delete() error {
...
+// the object should be deleted.
+func (builder *Builder) DeleteGraceful(gracePeriod *int64) error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	if *gracePeriod < int64(0) {
+		glog.V(100).Infof("GracePeriod cannot be non-negative integer %v", gracePeriod)
+
+		return fmt.Errorf("gracePeriod cannot be non-negative integer")
```
klaskosk, line 497: should be "cannot be negative" or "must be non-negative"

pkg/deployment/deployment.go
```diff
@@ -484,6 +484,45 @@ func (builder *Builder) Delete() error {
...
+	}
+
+	glog.V(100).Infof("Requested DeleteGracePeriod %v seconds", gracePeriod)
+
+	glog.V(100).Infof("Deleting deployment %s in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
```
klaskosk, line 503: Could you please combine these logs? Something like
```go
glog.V(100).Infof("Deleting deployment %s in namespace %s with %d second grace period", ...)
```

klaskosk (commented):
a couple of minor comments

pkg/deployment/deployment.go
yprokule: thanks, updated

pkg/deployment/deployment.go
yprokule: updated

pkg/deployment/deployment.go
```diff
@@ -484,6 +484,45 @@ func (builder *Builder) Delete() error {
...
+		return err
+	}
+
+	if *gracePeriod < int64(0) {
+		glog.V(100).Infof("gracePeriod(%v) must be non-negative integer", gracePeriod)
+
+		return fmt.Errorf("gracePeriod must be non-negative integer")
+	}
+
+	glog.V(100).Infof("Requested DeleteGracePeriod %v seconds", gracePeriod)
```
klaskosk, line 500: could you please remove this log now that the other one has been updated?

pkg/deployment/deployment.go
```diff
@@ -484,6 +484,45 @@ func (builder *Builder) Delete() error {
...
 }
 
+// DeleteGraceful removes a deployment while waiting for specified duration(in seconds)
+// the object should be deleted.
+func (builder *Builder) DeleteGraceful(gracePeriod *int64) error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	if *gracePeriod < int64(0) {
```
klaskosk, line 494: I just realized this will panic if gracePeriod is nil, maybe the parameter can be an int64 and then you take a reference later in the delete call? otherwise you'll need to make this `if gracePeriod != nil && *gracePeriod < int64(0)`

yprokule: good catch, updated

cdvultur (approved):
/lgtm

klaskosk (approved):
/lgtm

PR #998, opened 2025-04-15
PoC: Common Validate() func

sebrandon1:
Proof of concept for making a common interface for `Validate()` with `deployment` being the first package to take advantage of it.

### New validation mechanism:

* [`pkg/common/types.go`](diffhunk://#diff-2ff757ca5b194b9a983adc95868bd4b9600aa807cacfd2569801926e687a5ffcR1-R5): Added a new constant `DeploymentType`.
* [`pkg/common/validate.go`](diffhunk://#diff-38bda930798637d4f2d681cac4c5a43c9e647d0266f9f074380d3aed066cb583R1-R44): Introduced the `BuilderInterface` and the `ValidateBuilder` function to validate builder objects.
* [`pkg/common/validate_test.go`](diffhunk://#diff-909f495387af52986871255458cd43af0f0823e08707b828f4d639bf981fcbf9R1-R111): Added unit tests for the `ValidateBuilder` function using a mock builder.

### Integration with deployment package:

* [`pkg/deployment/deployment.go`](diffhunk://#diff-b44566f9caa976421532eb049c77b7eb247418f138516fab277a1ec33d3a7992L689-R714): Replaced the existing `validate` method with the new `ValidateBuilder` function and implemented the `BuilderInterface` methods.
* [`pkg/deployment/deployment_test.go`](diffhunk://#diff-e63ab873819d44774258d70d7498d2f000378f6611712dc641a356d19cb8b789L805-R818): Updated tests to reflect the changes in the validation mechanism.

### Import adjustments:

* [`pkg/deployment/deployment.go`](diffhunk://#diff-b44566f9caa976421532eb049c77b7eb247418f138516fab277a1ec33d3a7992L13-R13): Updated imports to include the new `common` package.
* [`pkg/deployment/deployment_test.go`](diffhunk://#diff-e63ab873819d44774258d70d7498d2f000378f6611712dc641a356d19cb8b789R10): Updated imports to include the new `common` package.

### DIFF
--- a/pkg/common/types.go
+++ b/pkg/common/types.go
@@ -1,0 +1,6 @@
+package common
+
+const (
+	// DeploymentType is the type for deployment resources.
+	DeploymentType = "deployment"
+)
--- a/pkg/common/validate.go
+++ b/pkg/common/validate.go
@@ -1,0 +1,50 @@
+package common
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+)
+
+// BuilderInterface is an interface that defining Builders.
+type BuilderInterface interface {
+	GetDefinition() interface{}
+	GetErrorMsg() string
+	GetAPIClient() interface{}
+	GetResourceType() string
+}
+
+// ValidateBuilder checks if the builder is valid.
+func ValidateBuilder(builder BuilderInterface) (bool, error) {
+	if builder == nil || reflect.ValueOf(builder).IsNil() {
+		glog.V(100).Info("The builder is uninitialized or nil")
+
+		return false, fmt.Errorf("error: received nil builder")
+	}
+
+	resourceType := builder.GetResourceType()
+
+	definition := builder.GetDefinition()
+	if definition == nil || (reflect.ValueOf(definition).Kind() == reflect.Ptr && reflect.ValueOf(definition).IsNil()) {
+		glog.V(100).Infof("The %s is undefined or has a nil underlying value", resourceType)
+
+		return false, fmt.Errorf("%s", msg.UndefinedCrdObjectErrString(resourceType))
+	}
+
+	apiClient := builder.GetAPIClient()
+	if apiClient == nil || (reflect.ValueOf(apiClient).Kind() == reflect.Ptr && reflect.ValueOf(apiClient).IsNil()) {
+		glog.V(100).Infof("The %s builder apiclient is nil", resourceType)
+
+		return false, fmt.Errorf("%s builder cannot have nil apiClient", resourceType)
+	}
+
+	if builder.GetErrorMsg() != "" {
+		glog.V(100).Infof("The %s builder has error message: %s", resourceType, builder.GetErrorMsg())
+
+		return false, fmt.Errorf("%s", builder.GetErrorMsg())
+	}
+
+	return true, nil
+}
--- a/pkg/common/validate.go
+++ b/pkg/common/validate.go
@@ -8,16 +8,16 @@ import (
 	"github.com/openshift-kni/eco-goinfra/pkg/msg"
 )
 
-// BuilderInterface is an interface that defining Builders.
-type BuilderInterface interface {
-	GetDefinition() interface{}
+// BuilderInterface is a generic interface defining Builders.
+type BuilderInterface[T any] interface {
+	GetDefinition() *T
 	GetErrorMsg() string
 	GetAPIClient() interface{}
 	GetResourceType() string
 }
 
 // ValidateBuilder checks if the builder is valid.
-func ValidateBuilder(builder BuilderInterface) (bool, error) {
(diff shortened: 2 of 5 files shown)


pkg/deployment/deployment.go
```diff
@@ -686,36 +686,37 @@ func GetGVR() schema.GroupVersionResource {
...
+// GetResourceType returns the resource type of the deployment.
+func (builder *Builder) GetResourceType() string {
+	return common.DeploymentType
+}
 
-	return true, nil
+// Replace the validate() method with a call to common.ValidateBuilder.
+func (builder *Builder) validate() (bool, error) {
+	return common.ValidateBuilder(builder)
 }
```
sebrandon1, line 720: I kept `builder.validate()` in there, but I'm more than willing to directly call `common.ValidateBuilder()` where necessary.

pkg/deployment/deployment_test.go
```diff
@@ -802,19 +803,19 @@ func TestValidate(t *testing.T) {
 			builderNil:    true,
 			definitionNil: false,
 			apiClientNil:  false,
-			expectedError: "error: received nil ClusterDeployment builder",
+			expectedError: "error: received nil builder",
```
sebrandon1, line 806: I'm also okay with just fully removing `deployment_test.go`'s version of TestValidate as well.  This is "extra" now.

pkg/common/validate.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+)
+
+// BuilderInterface is an interface that defining Builders.
+type BuilderInterface interface {
+	GetDefinition() interface{}
+	GetErrorMsg() string
+	GetAPIClient() interface{}
+	GetResourceType() string
+}
```
klaskosk, line 17: I wonder if we can do something like this so that the definition is typed. DeploymentBuilder will the implement `BuilderInterface[corev1.Deployment]` and then we could have `func ValidateBuilder[T any](builder BuilderInterface[T]) (bool, error)`
```go
type BuilderInterface[T any] interface {
	GetDefinition() *T
	GetErrorMsg() string
	GetAPIClient() any
	GetResourceType() string
}
```
have not tested this idea to see but it seems like a good way to incorporate generics

pkg/common/validate.go
```diff
@@ -0,0 +1,50 @@
...
+
+// ValidateBuilder checks if the builder is valid.
+func ValidateBuilder(builder BuilderInterface) (bool, error) {
```
klaskosk, line 20: since we've always written the bool so that it's `err == nil`, do we want to remove it now that we're going to refactor the function?

klaskosk, line 20: since we've always written the bool so that it's `err == nil`, do we want to remove it now that we're going to refactor the function?

pkg/common/validate.go
klaskosk: would also simplify since we don't really need reflect if we know that GetDefinition() returns a pointer to a type

pkg/common/validate.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+)
+
+// BuilderInterface is an interface that defining Builders.
+type BuilderInterface interface {
+	GetDefinition() interface{}
+	GetErrorMsg() string
+	GetAPIClient() interface{}
```
klaskosk, line 15: seeing this makes me wish we only have to worry about the controller-runtime client since it would make implementing functions like this for CRUD operations much easier

klaskosk (commented):
I've just added the thoughts that popped into my head, open questions/comments rather than requests for changes

pkg/common/validate.go
sebrandon1: I'll check that out.

sebrandon1: Updated in https://github.com/openshift-kni/eco-goinfra/pull/998/commits/68654e51731a2f48b7a2df4da98de937ca79ceb3

pkg/common/validate.go
sebrandon1: Done, removed.

pkg/common/validate.go
sebrandon1: I've got another branch for CRUD operations 😉

pkg/deployment/deployment.go
sebrandon1: Removed this func.

pkg/deployment/deployment_test.go
sebrandon1: Removed deployment TestValidate

pkg/common/types.go
```diff
@@ -0,0 +1,6 @@
+package common
+
+const (
+	// DeploymentType is the type for deployment resources.
+	DeploymentType = "deployment"
```
sebrandon1, line 5: Might be removed due to generics.

sebrandon1:
Closing in favor of: https://github.com/openshift-kni/eco-goinfra/pull/1008

PR #1003, opened 2025-04-18
poc: common runtime validate method

trewest:
- Create commonbuilder pkg
- Add public Validate method to be used by builders that embed commonbuilder
- Update clusterinstance to leverage embedded commonbuilder


internal/commonbuilder/commonbuilder_test.go
```diff
@@ -0,0 +1,100 @@
...
+			assert.True(t, result)
+		}
+	}
+}
+
+func generateTestBuilder() *testBuilder {
+	return &testBuilder{
+		apiClient: clients.GetTestClients(
+			clients.TestClientParams{}).Client,
+		Definition: generateClusterInstance(),
```
trewest, line 67: Used clusterinstance for this test so I wouldn't need to create my own struct that implemented everything required by runtime objects. The idea is that this shouldn't matter though, you could replace it with any pkg using runtime objects.

pkg/siteconfig/clusterinstance_test.go
```diff
@@ -821,72 +821,14 @@ func TestClusterInstanceExists(t *testing.T) {
...
-
-		if testCase.definitionNil {
-			testBuilder.Definition = nil
-		}
-
-		if testCase.apiClientNil {
-			testBuilder.apiClient = nil
-		}
+func generateClusterInstanceBuilderWithFakeObjects(objects []runtime.Object) *CIBuilder {
+	builder := NewCIBuilder(clients.GetTestClients(
```
trewest, line 825: This is one limitation of this approach. Since we initialize the commonbuilder only during NewBuilder and Pull functions, we have to generate test builders either through the NewBuilder function or manually initialize builder.CommonBuilder

internal/commonbuilder/types.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// CommonBuilderInterface is an interface implemented by all builders.
+type CommonBuilderInterface interface {
+	GetClient() goclient.Client
+	GetDefinition() goclient.Object
+	GetErrorMsg() string
+	GetKind() string
```
trewest, line 12: IMO, this one is a nice to have but not a deal breaker. It really only adds a little bit of context to the error messages but we could really let the user handle that in their own logging/error checking

achuzhoy (approved):

sebrandon1:
I like it.  That `goclient` is the real magic 🪄 basically replaces my use of generics.

pkg/siteconfig/clusterinstance_test.go
sebrandon1: For the manual step, could we create a `CreateTestCommbonBuilder` or something similar to the way we do `clients.GetTestClients`?

sebrandon1 (approved):

PR #1005, opened 2025-04-20
cnf network: fix nmstate ipv6 autoconfig

gkopels:
Hit flaky issue with ipv6 addresses:
[2025-04-20T08:34:54Z INFO  nmstate::query_apply::net_state] Rollbacked to checkpoint /org/freedesktop/NetworkManager/Checkpoint/33
NmstateError: NotImplementedError: Autoconf without DHCP is not supported yet

Added Autoconfig: false,


klaskosk (dismissed):
Since false is the zero value here it doesn't matter but it may be nice to set it explicitly in the test where you compare interfaces

klaskosk (approved):

sebrandon1 (approved):

PR #1008, opened 2025-04-22
PoC: common get/validate for builder

klaskosk:
I decided I'd try an attempt at a PoC, it draws a lot from Trey and Brandon's PoCs. Like Trey's it assumes that the controller-runtime client is the only one we'll use. I went with generics to allow us to return concrete types, but it has the weird workaround due to go's generics and getting the type without the pointer but still enforcing the interface on the pointer. 

Rust allows something like 
```rust
fn myFunc<T>(param: T) where for<'a> &'a T: MyInterface {}
```
but unfortunately this is not possible in a single type parameter in Go.

### DIFF
--- a/pkg/siteconfig/clusterinstance.go
+++ b/pkg/siteconfig/clusterinstance.go
@@ -718,6 +718,14 @@ func (builder *CIBuilder) GetClient() goclient.Client {
 	return builder.apiClient
 }
 
+func (builder *CIBuilder) SetClient(client goclient.Client) {
+	if builder == nil {
+		return
+	}
+
+	builder.apiClient = client
+}
+
 func (builder *CIBuilder) GetKind() schema.GroupVersionKind {
 	return siteconfigv1alpha1.GroupVersion.WithKind(siteconfigv1alpha1.ClusterInstanceKind)
 }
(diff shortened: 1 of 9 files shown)


pkg/siteconfig/clusterinstance.go
```diff
@@ -672,34 +652,85 @@ func (builder *CIBuilder) Exists() bool {
...
+func (builder *CIBuilder) Get() (*siteconfigv1alpha1.ClusterInstance, error) {
+	return common.Get(builder)
+}
+
+// validate will check that the builder and builder definition are properly initialized before
+// accessing any member fields.
+func (builder *CIBuilder) validate() (bool, error) {
+	err := common.Validate(builder)
+
+	return err == nil, err
```
trewest, line 735: Since the goal is to reduce new pkg/unit-test overhead, are we accomplishing that with this approach? I understand that we are offloading everything to a different pkg but still having these reside in the pkg will affect our code coverage, no?

klaskosk: That's right, these methods won't get unit tested themselves so the coverage percent will be affected. It doesn't really harm the actual coverage we have, only the number we generate

klaskosk:
Added an explanation for the PoC
[Rendered](https://github.com/klaskosk/eco-goinfra/blob/poc-common-interface/pkg/internal/common/README.md)

klaskosk:
Closing in favor of #1094 and subsequent PRs

PR #1020, opened 2025-05-01
ci: fix actions not using passed branch

klaskosk:
Workflows that are called by a scheduled workflow inherit the schedule event type so the check for workflow_call was always false. It is also unnecessary since the inputs will be null if not applicable so the `||` is enough to provide a fallback.


trewest (approved):
/lgtm

sebrandon1 (approved):

PR #1029, opened 2025-05-07
clusterlogging pkg update

elenagerman:
- ClusterLogging object handler removed as not supported since CLO 6.0
- modified ClusterLogForwarder api according to the changes were done for the CLO 6.0 +


pkg/clusterlogging/clusterlogforwarder.go
```diff
@@ -7,19 +7,19 @@ import (
...
 	"github.com/openshift-kni/eco-goinfra/pkg/msg"
-	clov1 "github.com/openshift/cluster-logging-operator/api/logging/v1"
+	observabilityv1 "github.com/openshift/cluster-logging-operator/api/observability/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	goclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// ClusterLogForwarderBuilder provides a struct for clusterlogforwarder object from the
+// Builder provides a struct for clusterlogforwarder object from the
```
klaskosk, line 16: pretty minor comment but I don't think there's a need to change the name of the builder since it's still not the only CR in CLO (even if LogFileExporter is a different API group I think it'd make sense to have in this package in the future)
feel free to disagree though since `ClusterLogForwarderBuilder` is a long name

pkg/clusterlogging/clusterlogforwarder.go
```diff
@@ -74,8 +74,48 @@ func NewClusterLogForwarderBuilder(
 	return builder
 }
 
+// WithManagementState sets the clusterlogforwarder operator's managementState configuration.
+func (builder *Builder) WithManagementState(
+	managementState observabilityv1.ManagementState) *Builder {
```
klaskosk, line 79: since `observabilityv1.ManagementState` is an enum, could you please validate that the function argument is a valid management state?

pkg/clusterlogging/clusterlogforwarder_test.go
```diff
@@ -316,27 +320,82 @@ func TestClusterLogForwarderUpdate(t *testing.T) {
...
 		} else {
-			assert.Equal(t, []clov1.PipelineSpec{testCase.pipelines},
+			assert.Equal(t, []observabilityv1.PipelineSpec{testCase.pipelines},
 				testCase.testClusterLogForwarder.Definition.Spec.Pipelines)
 		}
 	}
 }
 
+func TestWithManagementState(t *testing.T) {
+	testCases := []struct {
```
klaskosk, line 330: related to the above, could you please add a case for an invalid management state here?

klaskosk (commented):
thanks for updating this for CLO 6! just a few minor comments

pkg/clusterlogging/clusterlogforwarder.go
elenagerman: sure, np at all, I change it back; thank you!

pkg/clusterlogging/clusterlogforwarder.go
elenagerman: I can add this verification, but there is no way to test it and and don't see any valuable valu for this verification, sorry.
There are only two  options and the input variable type defined as observabilityv1.ManagementState, so there is no any way we can define any except of this two.
Please clarify if I didn't understand it correctly

pkg/clusterlogging/clusterlogforwarder_test.go
elenagerman: for the same reason I cannot. the only option is empty variable maybe.

pkg/clusterlogging/clusterlogforwarder.go
klaskosk: @elenagerman I just meant add a check something like this:
```go
if managementState != observabilityv1.ManagementStateManaged && managementState != observabilityv1.ManagementStateUnmanaged {
    builder.errorMsg = fmt.Sprintf("management state %s is invalid", managementState)
}
```
Since you could provide it with a string that's converted into the management state (or any string literal I think) and it could be invalid. It's not that big a deal since you'd have to be doing something incorrect to get into that situation but it's nice to ensure the With methods only result in valid resources

klaskosk (approved):
/lgtm

sebrandon1 (approved):

PR #1039, opened 2025-05-13
argocd: add git functions to application

klaskosk:
As part of an effort to decouple and make reusable the gitdetails helper from cnf/ran in eco-gotests, this PR adds more generic versions of these helpers as methods on the Application builder itself.

WithGitPathAppended is a more specific version of WithGitDetails that is more aligned with the common use case of updating to use a subdirectory of the git repo. WithGitDetails may still be used to update all of the git details if necessary.

DoesGitPathExist can be called similar to WithGitPathAppended and checks if that path would exist. The way it checks should be mostly portable between git hosts, at least enough to cover the common cases.

WaitForSourceUpdate allows verifying a change to the Application's source has taken effect.

### DIFF
--- a/pkg/argocd/applications.go
+++ b/pkg/argocd/applications.go
@@ -2,7 +2,12 @@ package argocd
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


klaskosk:
cc: @josclark42

sebrandon1 (approved):

PR #1044, opened 2025-05-21
daemonset: IsReady poll immediate set to false

ajaggapa:
Setting `PollUntilContextTimeout` argument `immediate` to `false` to avoid racy scenario leading to test failure


klaskosk (commented):
What's the reason that setting immediate to `true` causes a race? If there's something about the resource that makes it necessary (DesiredNumberScheduled is zero for a few seconds or so, for example) it should be documented to explain why it'd need to break with consistency

ajaggapa:
@klaskosk 

We have a scenario where the daemonset is recreated and we wait for it to be Ready. 
Polling function is returning error if the Ds is not present in the cluster and thus the `IsReady` function is terminated without any retries.

Below log is when i bypassed returning error inside poll.
```
18:59:58  I0521 13:29:57.896316   22865 daemonset.go:434] Running periodic check until daemonset speaker in namespace metallb-system is ready or timeout 3m0s exceeded
18:59:58  I0521 13:29:57.896340   22865 daemonset.go:418] Checking if daemonset speaker exists in namespace metallb-system
18:59:58  I0521 13:29:57.906484   22865 daemonset.go:450] Failed to get daemonset from cluster. Error is: 'daemonsets.apps "speaker" not found'
19:00:01  I0521 13:30:00.898680   22865 daemonset.go:418] Checking if daemonset speaker exists in namespace metallb-system
```


Solution1:
- Set immediate to false
With this we just a wait for 1 retry interval (3s) for the Ds to be recreated. 

Solution2:
- Do not return error inside poll function
I have now pushed changes to include this solution instead of the above. Now `IsReady` waits for the timeout to complete even if the Ds is not present in the cluster.

klaskosk (approved):
Thanks for explaining further! I prefer this solution since it is more consistent with other packages. Really you could get rid of the entire if statement since Exists is calling Get in a similar way, but I'm fine approving as is

klaskosk (approved):

trewest (approved):
/lgtm

PR #1051, opened 2025-06-04
Added WaitForReinstallCondition function to siteconfig

josclark42:
Adds a function to siteconfig to wait for Status.Reinstall.Conditions.

### DIFF
--- a/pkg/siteconfig/clusterinstance.go
+++ b/pkg/siteconfig/clusterinstance.go
@@ -491,6 +491,70 @@ func (builder *CIBuilder) WaitForCondition(expected metav1.Condition, timeout ti
 	return builder, err
 }
 
+// WaitForReinstallCondition waits until the ClusterInstance
+// has a reinstall condition that matches the expected, checking only the Type, Status, Reason, and Message fields.
+// For the message field, it matches if the message contains the expected.
+// Zero fields in the expected condition are ignored.
+func (builder *CIBuilder) WaitForReinstallCondition(expected metav1.Condition,
+	timeout time.Duration) (*CIBuilder, error) {
+	if valid, err := builder.validate(); !valid {
+		return builder, err
+	}
+
+	if !builder.Exists() {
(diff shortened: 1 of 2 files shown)


pkg/siteconfig/clusterinstance.go
```diff
@@ -491,6 +491,70 @@ func (builder *CIBuilder) WaitForCondition(expected metav1.Condition, timeout ti
...
+
+			if err != nil {
+				glog.V(100).Info("failed to get clusterinstance %s/%s: %w",
+					builder.Definition.Namespace, builder.Definition.Name, err)
+
+				return false, nil
+			}
+
+			if builder.Object.Status.Reinstall == nil {
+				glog.V(100).Info("failed to get reinstall status %s/%s: %w",
```
klaskosk, line 524: this may have been wrong in the other function in fairness, but these should be `%v` instead of `%w`

`%w` should only be used inside `fmt.Errorf` since it has a special meaning of wrapping an error. `%v` should be used where you just want to print the string representation of an error

josclark42: changed- thanks for explaining.

pkg/siteconfig/clusterinstance.go
```diff
@@ -491,6 +491,70 @@ func (builder *CIBuilder) WaitForCondition(expected metav1.Condition, timeout ti
...
+			builder.Object, err = builder.Get()
+
+			if err != nil {
+				glog.V(100).Info("failed to get clusterinstance %s/%s: %v",
+					builder.Definition.Namespace, builder.Definition.Name, err)
+
+				return false, nil
+			}
+
+			if builder.Object.Status.Reinstall == nil {
```
trewest, line 523:
```suggestion
			if builder.Object.Status.Reinstall == nil || clusterInstace.Object.Status.Reinstall.Conditions == nil {
```

josclark42: Added check for Reinstall.Conditions. Thanks.

PR #1056, opened 2025-06-13
hw-accel: prepare remove of preflightvalidationocp crd

cdvultur:
The new `release-2.4`  of [KMM](https://github.com/rh-ecosystem-edge/kernel-module-management/tree/release-2.4/api/v1beta2) will drop the `preflightvalidationocp` crd. 

The tests will be moved to the u/s crd: `preflightvalidation`

This PR is a precursor of the crd removal and assures that there are no breakages.

### DIFF
--- a/pkg/kmm/preflightvalidation.go
+++ b/pkg/kmm/preflightvalidation.go
@@ -1,0 +1,327 @@
+package kmm
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	kmmv1beta2 "github.com/openshift-kni/eco-goinfra/pkg/schemes/kmm/v1beta2"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	goclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// PreflightValidationBuilder provides struct for the preflightvalidation object
+// containing connection to the cluster and the preflightvalidation definitions.
+type PreflightValidationBuilder struct {
+	// PreflightValidation definition. Used to create a PreflightValidation object.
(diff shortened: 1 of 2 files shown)


pkg/kmm/preflightvalidation.go
```diff
@@ -0,0 +1,327 @@
+package kmm
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
+	"github.com/openshift-kni/eco-goinfra/pkg/clients"
+	"github.com/openshift-kni/eco-goinfra/pkg/msg"
+	kmmV1Beta2 "github.com/openshift-kni/eco-goinfra/pkg/schemes/kmm/v1beta2"
```
klaskosk, line 10: nit: the more consistent capitalization would be `kmmv1beta2`

pkg/kmm/preflightvalidation.go
```diff
@@ -0,0 +1,327 @@
...
+		return false
+	}
+
+	glog.V(100).Infof("Checking if preflightvalidation %s exists in namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	var err error
+	builder.Object, err = builder.Get()
+
+	return err == nil || !k8serrors.IsNotFound(err)
```
klaskosk, line 241: this may cause issues to have exists return true for other errors ("the Exists issue") but since we're going to fix that within the 4.20 timeframe anyway not a huge deal for this PR

I'd recommend logging the error in `Get` so it's not a silent failure but if it's not something you run into it's not a huge deal

klaskosk (approved):
some nits and a warning about the `Exists` method, but overall looks good and we'll end up resolving the warning with the common methods next release anyway

pkg/kmm/preflightvalidation.go
cdvultur: Fixed it. Thanks!

pkg/kmm/preflightvalidation.go
cdvultur: Ok. I'll leave it like that and sort it out in all the `Exists` methods inside `kmm` pkg in a future PR

PR #1062, opened 2025-06-24
oran api: add client for artifacts

klaskosk:
This PR adds an additional API client to the oran/api package for the artifacts. Currently, the only artifact is the ManagedInfrastructureTemplate, which corresponds with the ClusterTemplate resource. Unlike with the ProvisioningClient, the ArtifactsClient does not work as a runtimeclient.Client that can be used with the ClusterTemplateBuilder. This is because there are only three different operations that can be performed. Additionally, filtering and defaults are not easily accessible with the existing Builder API.

Also included in this PR is the filter package in oran/api. This corresponds to the filtering common to all APIs and provides a structured way to create filter expressions.


cdvultur (approved):
/lgtm for what is outside `internal/`

sebrandon1 (approved):

PR #1076, opened 2025-07-17
pkg metallb: Added v1beta2 API

evgenLevin:


pkg/metallb/mlbtypesv1beta2/bgppeer_types.go
```diff
@@ -0,0 +1,160 @@
+/*
```
klaskosk, line 1: I think you can just add a nolint on this file since it's not really our code
```go
//nolint // Code taken from another repo so no need to apply our linter
```

klaskosk (commented):
Just need to get rid of the linter errors otherwise lgtm

pkg/metallb/mlbtypesv1beta2/bgppeer_types.go
```diff
@@ -0,0 +1,163 @@
+//nolint
```
klaskosk, line 1: if this doesn't work there should be a way to exclude linter errors from this path in `.golangci.yml`

PR #1094, opened 2025-08-07
common: initial interface and read-only functions

klaskosk:
This PR provides the groundwork for the refactoring of builders based on the PoC in #1008. As the initial PR, it does not include the embeddable builder and instead focuses on the common interface and the methods required to cover read-only cluster-scoped and namespaced builders. A notable change from the PoC is the inclusion of separate logging and errors files that are intended to make it easier to refactor these areas in the future.

Some tweaks to the linting configuration have also been made, which is why pkg/oran/api has been updated.

Assisted-by: Cursor


sebrandon1 (dismissed):

klaskosk:
@coderabbitai review

pkg/internal/common/common.go
```diff
@@ -0,0 +1,328 @@
...
+// NewClusterScopedBuilder creates a new builder for a cluster-scoped resource. It is generic over the actual builder
+// type and uses the methods from the Builder interface to create the actual builder. Generic parameters are ordered so
+// that SO and SB can be elided and only O and B must be provided.
+func NewClusterScopedBuilder[O, B any, SO objectPointer[O], SB builderPointer[B, O, SO]](
+	apiClient runtimeclient.Client, schemeAttacher clients.SchemeAttacher, name string) SB {
+	var builder SB = new(B)
+
+	logNewClusterScopedBuilderInitializing(builder.GetGVK().Kind, name)
+
+	if apiClient == nil || reflect.ValueOf(apiClient).IsNil() {
```
trewest, line 81: Can we create a common function for this check?

trewest (changes requested):
Just one question, otherwise lgtm

klaskosk:
@coderabbitai resolve

PR #1101, opened 2025-08-08
pod: attempt to address integration test flakiness

klaskosk:
This PR attempts to address the cause behind the integration test flakiness by not redirecting stdin in ExecCommand.

Assisted-by: Cursor


klaskosk:
I've run the integration tests 5 times with this change and all of them succeeded on the first try. `eco-gotests` does not take advantage of the stdin anywhere (and in Jenkins pipelines certainly won't) so I think we're clear to remove the `Stdin` field. If anything it should improve reliability a bit

sebrandon1 (approved):

PR #1136, opened 2025-09-07
Add HasTag function to check imagestream tags

ItzikEzra-rh:
- Add HasTag function that checks  tags in imagestream

## Summary by CodeRabbit

* **New Features**
  * Retrieve the current list of published tags for an image stream from its status.

* **Bug Fixes / Reliability**
  * Returns a clear error when the image stream does not exist.
  * Returns a descriptive error when an image stream has no status tags; improved diagnostic logging for tag retrieval.

* **Tests**
  * Added unit test covering successful tag retrieval and the empty-status error case.

### DIFF
--- a/pkg/imagestream/imagestream.go
+++ b/pkg/imagestream/imagestream.go
@@ -166,12 +166,11 @@ func (builder *Builder) GetStatusTags() ([]string, error) {
 		return nil, err
 	}
 
+	if !builder.Exists() {
+		glog.V(100).Infof("ImageStream %s does not exist", builder.Definition.Name)
 
-    if !builder.Exists() {
-        glog.V(100).Infof("ImageStream %s does not exist",  builder.Definition.Name)
-        
-        return nil, fmt.Errorf("cannot get status tags from non-existent imageStream")
-    }
+		return nil, fmt.Errorf("cannot get status tags from non-existent imageStream")
+	}
 
 	if len(builder.Object.Status.Tags) == 0 {
 		return nil, fmt.Errorf("imageStream object %s in namespace %s has no status tags",


pkg/imagestream/imagestream.go
```diff
@@ -160,6 +160,46 @@ func (builder *Builder) GetDockerImage(imageTag string) (string, error) {
 		imageTag, builder.Definition.Name, builder.Definition.Namespace)
 }
 
+// HasTag checks if the imageStream has a specific tag.
+func (builder *Builder) HasTag(tagName string) (bool, error) {
```
cdvultur, line 164: I think that the HasTag checks can be done with GetDockerImage . If the tag does not exist , it will give same message. 

I  would try to use that function, because it covers the same functionality

cdvultur: In fact, the `GetDockerImage` checks only the .spec. In our case we need to check the .status for the ImageStream. 

In this case just rename the function for `HasTagInStatus` so that it is clear that the check is in different branch

cdvultur (approved):

pkg/imagestream/imagestream_test.go
```diff
@@ -264,6 +264,34 @@ func TestImageStreamGetDockerImage(t *testing.T) {
 	}
 }
 
+func TestImageStreamGetStatusTags(t *testing.T) {
```
trewest, line 267: Can you also add a negative test?

ItzikEzra-rh: sure

trewest (dismissed):
/lgtm

PR #1143, opened 2025-09-16
Method ExecCommandWithTimeout added

yprokule:
Assisted-by: Cursor

## Summary by CodeRabbit

- **New Features**
  - Added ability to run commands inside a pod with a configurable timeout, container selection, and enhanced execution logging.

- **Bug Fixes**
  - Fail early with a clear error when the target pod does not exist.

- **Tests**
  - Added tests covering timeout execution, input validation, container selection behavior, and non-existent pod handling.

### DIFF
--- a/pkg/pod/pod.go
+++ b/pkg/pod/pod.go
@@ -428,6 +428,14 @@ func (builder *Builder) ExecCommand(command []string, containerName ...string) (
 		return bytes.Buffer{}, err
 	}
 
+	if !builder.Exists() {
+		glog.V(100).Infof("Cannot execute command on pod %s in namespace %s because it does not exist",
+			builder.Definition.Name, builder.Definition.Namespace)
+
+		return bytes.Buffer{}, fmt.Errorf("pod object %s does not exist in namespace %s",
+			builder.Definition.Name, builder.Definition.Namespace)
+	}
+
 	var (
 		buffer bytes.Buffer
 		cName  string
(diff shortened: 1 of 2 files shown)


pkg/pod/pod.go
```diff
@@ -475,6 +483,90 @@ func (builder *Builder) ExecCommand(command []string, containerName ...string) (
...
+		cName  string
+	)
+
+	if len(containerName) > 0 {
+		cName = containerName[0]
+	} else {
+		cName = builder.Definition.Spec.Containers[0].Name
+	}
+
+	glog.V(100).Infof("Execute command %v in the pod %s container %s in namespace %s with %s timeout",
```
klaskosk, line 531: resolved the coderabbit since its solution was wrong, but please do update this function to use `builder.Definition` instead of `builder.Object` since it avoids a class of panics that could occur

trewest (approved):
/lgtm

pkg/pod/pod_test.go
```diff
@@ -649,6 +650,177 @@ func TestPodIsHealthy(t *testing.T) {
...
+		{
+			name:          "no container specified - uses default",
+			containerName: []string{},
+		},
+		{
+			name:          "container specified",
+			containerName: []string{"custom-container"},
+		},
+		{
+			name:          "multiple containers specified - uses first",
```
klaskosk, line 743: not that important but this doesn't actually test that it uses the first, just that it doesn't cause an error by providing multiple containers

pkg/pod/pod_test.go
```diff
@@ -649,6 +650,177 @@ func TestPodIsHealthy(t *testing.T) {
...
+
+			// Should get "does not exist" error, not a parameter validation error
+			assert.Error(t, err)
+			assert.Contains(t, err.Error(), "does not exist in namespace")
+		})
+	}
+}
+
+// TestPodExecCommandWithTimeoutEdgeCases tests edge cases and boundary conditions.
+func TestPodExecCommandWithTimeoutEdgeCases(t *testing.T) {
```
klaskosk, line 771: it doesn't really hurt to have but this test doesn't expand the coverage of the function. Also the single character command is two characters 🙂

pkg/pod/pod.go
yprokule: `builder.Exists()` on line 512 sets `builder.Object` or exits if it doesn't, so this should be enough to guard against a class of panics that could occur

pkg/pod/pod_test.go
yprokule: thanks, removed

PR #1161, opened 2025-10-02
ingress: add new builder for Ingress resource

klaskosk:
The existing ingress package only contains a builder for the IngressController, not the Ingress. This PR fills this gap by adding a new builder for the Ingress resource.

Assisted-by: Cursor


pkg/ingress/ingress.go
```diff
@@ -0,0 +1,286 @@
...
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	ingress := &networkingv1.Ingress{}
+
+	err := builder.apiClient.Get(context.TODO(), goclient.ObjectKey{
+		Name:      builder.Definition.Name,
+		Namespace: builder.Definition.Namespace,
+	}, ingress)
+	if err != nil {
+		return nil, err
```
yprokule, line 148: worth logging an error for further troubleshooting

pkg/ingress/ingress.go
```diff
@@ -0,0 +1,286 @@
...
+			builder.Definition.Name, builder.Definition.Namespace)
+
+		builder.Object = builder.Definition
+
+		return builder, nil
+	}
+
+	err := builder.apiClient.Create(context.TODO(), builder.Definition)
+	if err != nil {
+		return nil, err
```
yprokule, line 195: ditto, worth logging an error for further troubleshooting

pkg/ingress/ingress.go
```diff
@@ -0,0 +1,286 @@
...
+
+	if !builder.Exists() {
+		builder.Object = nil
+
+		return nil
+	}
+
+	err := builder.apiClient.Delete(context.TODO(), builder.Definition)
+	if err != nil {
+		return fmt.Errorf("cannot delete ingress: %w", err)
```
yprokule, line 248: ditto

pkg/ingress/ingress.go
```diff
@@ -0,0 +1,286 @@
...
+func (builder *IngressBuilder) Delete() error {
+	if valid, err := builder.validate(); !valid {
+		return err
+	}
+
+	glog.V(100).Infof("Deleting the ingress %s from namespace %s",
+		builder.Definition.Name, builder.Definition.Namespace)
+
+	if !builder.Exists() {
+		builder.Object = nil
```
yprokule, line 241: I'd probably log this as well

pkg/ingress/ingress.go
```diff
@@ -0,0 +1,301 @@
+package ingress
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/golang/glog"
```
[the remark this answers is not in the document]
klaskosk: @sebrandon1 do we need to make changes to eco-gotests to ensure the flags (mainly `logtostderr`) are set correctly? If not, I can switch this file over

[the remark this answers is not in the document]
sebrandon1: I don't believe that is needed but I could be wrong 😄 

I did the same changes to our repo that also uses eco-goinfra and I didn't have to make changes other than just swap glog for klog: https://github.com/redhat-best-practices-for-k8s/certsuite-qe/pull/1215

PR #1327, opened 2026-04-16
network-operator: update setmultinetpolicy

gkopels:
Fix SetMultiNetworkPolicy(false, …) to avoid waiting for Progressing=True, which often never appears when disabling. On disable, update the CR then wait for Progressing=False (with the provided timeout) and Available=True (60s); enable behavior remains unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved multi-network policy toggling to wait for appropriate readiness states when enabling or disabling, reducing flakiness and unnecessary delays.
  * Uses the provided timeout for availability checks instead of a fixed short interval.
* **Style/Logging**
  * Enhanced operator wait logs to show expected condition/status and last observed state on timeout.
* **Chores**
  * Removed the IP forwarding configuration option.

### DIFF
--- a/pkg/network/operator.go
+++ b/pkg/network/operator.go
@@ -2,6 +2,7 @@ package network
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
@@ -203,7 +204,13 @@ func (builder *OperatorBuilder) SetIPForwarding(
 	return builder, err
 }
 
-// SetMultiNetworkPolicy enables network.operator multinetworkpolicy feature.
+// SetMultiNetworkPolicy enables or disables the network.operator multinetworkpolicy feature.
+// When disabling, the Cluster Network Operator often does not set Progressing=True; the initial
+// wait for Progressing=True is skipped in that case only. Enabling keeps the original sequence.
+// On disable, we wait until status.observedGeneration catches metadata.generation from the update
+// so condition polling is not satisfied from stale status from before this change.
+// The wait for Progressing=False uses a disable-specific path: CNO may omit the Progressing
+// condition when idle, which would otherwise never match WaitUntilInCondition(..., False).
 func (builder *OperatorBuilder) SetMultiNetworkPolicy(state bool, timeout time.Duration) (*OperatorBuilder, error) {
 	if valid, err := builder.validate(); !valid {
 		return builder, err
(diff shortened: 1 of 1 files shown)


pkg/network/operator.go
```diff
@@ -221,10 +221,12 @@ func (builder *OperatorBuilder) SetMultiNetworkPolicy(state bool, timeout time.D
...
-		err = builder.WaitUntilInCondition(
-			operatorv1.OperatorStatusTypeProgressing, 60*time.Second, operatorv1.ConditionTrue)
-		if err != nil {
-			return nil, err
+		if state {
+			err = builder.WaitUntilInCondition(
+				operatorv1.OperatorStatusTypeProgressing, 60*time.Second, operatorv1.ConditionTrue)
+			if err != nil {
+				return nil, err
+			}
```
evgenLevin, line 229: Maybe using a generation count is a WA. WDYT?
```suggestion

		// Capture the generation that the API server assigned to this update.
		// We need it to know when the operator has actually started processing *our* change
		// (as opposed to still reflecting a previous reconciliation).
		targetGeneration := builder.Definition.Generation
		if state {
			err = builder.WaitUntilInCondition(
				operatorv1.OperatorStatusTypeProgressing, 60*time.Second, operatorv1.ConditionTrue)
			if err != nil {
				return nil, err
			}
		} else {
			// When disabling, Progressing=True may never appear.
			// Instead, wait until observedGeneration catches up to the generation
			// we just wrote — that proves the operator has seen and processed our change.
			err = builder.waitUntilObservedGeneration(targetGeneration, timeout)
			if err != nil {
				return nil, err
			}
		}
		......
		
		
		// waitUntilObservedGeneration waits until status.observedGeneration >= the target generation,
// proving the operator has reconciled the spec change.
func (builder *OperatorBuilder) waitUntilObservedGeneration(
	targetGeneration int64, timeout time.Duration) error {
	if valid, err := builder.validate(); !valid {
		return err
	}
	klog.V(100).Infof("Waiting for network.operator %s observedGeneration >= %d",
		builder.Definition.Name, targetGeneration)
	return wait.PollUntilContextTimeout(
		context.TODO(), 3*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
			if !builder.Exists() {
				return false, fmt.Errorf("network.operator object %s does not exist", builder.Definition.Name)
			}
			return builder.Object.Status.ObservedGeneration >= targetGeneration, nil
		})
}
		
			
```

gkopels: Good idea — waiting for observedGeneration >= generation after Update() can anchor waits on our spec change. I’d only add it after confirming CNO bumps observedGeneration reliably for useMultiNetworkPolicy on our versions. This PR stays minimal: skip Progressing=True on disable, keep Progressing=False + Available=True.

klaskosk (commented):
I think Evgeny has a point, it's possible that we send off these requests before the reconciliation and end up returning from the function before the change was applied

but ultimately I'll defer to whatever y'all decide on for the behavior itself, the code itself is fine

gkopels:
> I think Evgeny has a point, it's possible that we send off these requests before the reconciliation and end up returning from the function before the change was applied
> 
> but ultimately I'll defer to whatever y'all decide on for the behavior itself, the code itself is fine

Hi ok I get the point and agree. I made a change and tested it. Please have a look Thanks!

pkg/network/operator.go
```diff
@@ -306,9 +279,85 @@ func (builder *OperatorBuilder) WaitUntilInCondition(
 			return false, nil
 		})
 
+	if err != nil && errors.Is(err, context.DeadlineExceeded) && builder.Exists() {
+		klog.V(100).Infof("timeout waiting for network.operator %s condition %s=%s; last status conditions: %#v",
+			builder.Definition.Name, condition, status, builder.Object.Status.Conditions)
+	}
```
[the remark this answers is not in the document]
gkopels: updated

klaskosk (approved):
lgtm, it'd be good to have Evgeny re-review too

gkopels:
> SetMultiNetworkPolicy

gkopels:
> lgtm, it'd be good to have Evgeny re-review too

Thanks Kirstin asking Evgeny to take a look.

evgenLevin (approved):
/lgtm

trewest (approved):

PR #1328, opened 2026-04-16
common: add discard logger to client operations

klaskosk:
This commit adds the discard logger to the context that gets passed to
the individual runtime client methods. It gets added only at this lower
level so that we do not miss our own logs coming from the common
package.

Assisted-by: Cursor


klaskosk:
Merging so we unblock the eco-goinfra bumps in eco-gotests

PR #1342, opened 2026-05-01
lca ipc: backporting new functions from main

achuzhoy:
The added functions were missing for some new
tests.


achuzhoy:
cc @trewest

PR #1497, opened 2026-05-28
common: add helper to convert metav1 to runtimeclient ListOptions

klaskosk:
This commit adds a helper for converting metav1.ListOptions into
runtimeclient.ListOptions. Builders for core resources like configmap or
namespace did not use the controller-runtime client prior to the common
builder migration, so their List functions used the metav1 options
rather than the runtimeclient ones. To keep the signatures the same, we
have to introduce a conversion helper.

The only risk here is for callers which used options that are not
supported by the controller-runtime client. These options relate to
watching and a timeout. However, the only places they are used in
eco-gotests they behave as no-ops, so this change will not impact them.

Depends-on: #1496 

Assisted-by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved listing behavior for namespaces and config maps, including better handling of empty namespace input and selector-based filtering.
  * Made list operations more consistent across views, with support for additional list parameters like limits and continuation tokens.

* **Refactor**
  * Streamlined list retrieval to use shared logic, which should improve consistency and reduce duplicate behavior.

* **Tests**
  * Updated coverage for listing flows and selector conversion scenarios, including invalid selector handling.

### DIFF
--- a/pkg/configmap/list.go
+++ b/pkg/configmap/list.go
@@ -1,112 +1,45 @@
 package configmap
 
 import (
-	"fmt"
+	"context"
 
 	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/clients"
-	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/internal/logging"
+	"github.com/rh-ecosystem-edge/eco-goinfra/pkg/internal/common"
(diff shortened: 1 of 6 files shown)


klaskosk:
merging as part of release-4.22 branching

PR #1502, opened 2026-06-01
oran api: normalize O2IMS provisioning phase casing and sync OpenAPI spec

josclark42:
After oran-o2ims aligned with O-RAN R005-v11, the provisioning API
returns uppercase phase values (e.g. FAILED) while Kubernetes CRs use
lowercase. Normalize phases when mapping API responses to CR objects so
WaitForPhaseAfter matches correctly. Resync the provisioning OpenAPI
definitions with upstream and regenerate the client for related schema
changes (provisionedResourceSet rename, new status fields, delete 202).

PR that introduced uppercase FAILED
[openshift-kni/oran-o2ims #2369](https://github.com/openshift-kni/oran-o2ims/pull/2369)

## Summary by CodeRabbit

* **New Features**
  * POST and DELETE provisioning responses include Location headers with URI references.
  * Provisioning status now exposes detailed per-resource and per-node-cluster provisioning status.

* **Chores**
  * API contract version bumped to 1.2.0.
  * DELETE provisioning request now returns HTTP 202 Accepted.

* **Tests**
  * Added unit test verifying provisioning phase normalization from uppercase to lowercase.


pkg/oran/api/internal/provisioning/openapi.yaml
klaskosk, line 1: 1 minor thing, could you please add a comment to show what commit this openapi is from? I started it when I added the alarms client (that's why it's missing here 🙂), that way we have some form of versioning

```
# Synced with oran-o2ims commit <commit sha>
```

pkg/oran/api/provisioning.go
```diff
@@ -344,12 +345,14 @@ func provisioningRequestFromInfo(
 		},
 		Status: provisioningv1alpha1.ProvisioningRequestStatus{
 			ProvisioningStatus: provisioningv1alpha1.ProvisioningStatus{
-				ProvisioningPhase:   provisioningv1alpha1.ProvisioningPhase(unwrapOrDefault(info.Status.ProvisioningPhase)),
-				ProvisioningDetails: unwrapOrDefault(info.Status.Message),
+				ProvisioningPhase: provisioningv1alpha1.ProvisioningPhase(
+					strings.ToLower(string(info.Status.ProvisioningPhase)),
```
klaskosk, line 349: since we're following a spec update and the spec uses upper case, could you please remove the `strings.ToLower`? I'm hoping eco-gotests doesn't reference the string (and instead uses the constant), but I'd rather we update the tests to match the spec if we have to

josclark42: I think we need to normalize the string ToLower here because there is a case mismatch between the O2IMS REST response (upper), and [the type defined on the kubernetes side](https://github.com/rh-ecosystem-edge/eco-goinfra/blob/main/vendor/github.com/openshift-kni/oran-o2ims/api/provisioning/v1alpha1/provisioningrequest_types.go#L111-L134) (lower) .  I don't think it can be done on the eco-gotests side because we're not doing a string comparison, we're just waiting for StateFailed.

pkg/oran/api/internal/provisioning/openapi.yaml
josclark42: Done! Thanks.

pkg/oran/api/provisioning.go
klaskosk: ah good catch. That should be fixed when we update the go.mod dependencies, but we can do that later

klaskosk (approved):

PR #1505, opened 2026-06-03
common,ptp: broaden ListInAllNamespacesFunc

klaskosk:
Previously, the ListInAllNamespacesFunc only represented types which
accepted the runtimeclient.ListOptions struct. However, the
ListHardwareConfigs function in the ptp package accepts
runtimeclient.ListOption, which is the interface that
*runtimeclient.ListOptions implements.

This commit solves the issue by adding an extra generic parameter to the
ListInAllNamespacesFunc type which allows for specifying the options
type. That way, both forms may be provided to NewListTestConfig.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Tests**
  * Updated shared test utilities for list test configuration to support caller-supplied list option types, enabling more flexible generic list handling.
  * Simplified list-test wiring in hardware configuration/profile tests by directly passing the list functions instead of wrapper logic.

### DIFF
--- a/pkg/internal/common/testhelper/listhelper.go
+++ b/pkg/internal/common/testhelper/listhelper.go
@@ -16,10 +16,15 @@ import (
 )
 
 // ListInAllNamespacesFunc is a List function signature for listing resources in all namespaces (e.g.,
-// ListInAllNamespaces).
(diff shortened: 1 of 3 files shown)


klaskosk:
merging as part of release-4.22 branching

PR #1519, opened 2026-06-16
pod: add PortForward method to pod.Builder

frajamomo:
Wraps `k8s.io/client-go/tools/portforward` to provide a simple API for establishing port-forwards to pods. Returns the local address and a stop function, following the same validation pattern as ExecCommand.

Adds `PortForward(localPort, remotePort int) (string, func(), error)` to `pod.Builder`
Wraps k8s.io/client-go/tools/portforward with the standard builder validation pattern (same as `ExecCommand`)
Returns local address and a stop function for clean teardown

**Motivation**
Test suites that validate TLS security profiles need to probe pod endpoints via port-forwarding. Without this wrapper, each test suite must directly import `k8s.io/client-go/tools/portforward` and `transport/spdy`, duplicating ~40 lines of SPDY setup boilerplate.

This follows the same pattern as `ExecCommand`, which wraps remotecommand — both are pod operations that require SPDY transport setup.

Test plan

-  go build ./pkg/pod/... passes
-  go vet ./pkg/pod/... passes
-  Verified in eco-gotests CAPOA TLS profile test suite (PR rh-ecosystem-edge/eco-gotests#1383)

## Summary by CodeRabbit

* **New Features**
  * Added pod port-forwarding support to connect local ports to remote pod ports, including startup readiness coordination, a setup timeout, and an explicit stop handle to end the session cleanly.
* **Internal Improvements**
  * Improved the underlying port-forward transport to support WebSocket/stream tunneling and added a fallback dialing mechanism for more resilient connection establishment.

### DIFF
--- a/pkg/pod/portforward.go
+++ b/pkg/pod/portforward.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"sync"
 	"time"
 
 	"k8s.io/client-go/tools/portforward"
@@ -72,9 +73,18 @@ func (builder *Builder) PortForward(localPort, remotePort int) (string, func(), 
 			builder.Object.Name, defaultPortForwardReadyTimeout)
 	}
 
-	stop := func() {
+	ports, err := forwarder.GetPorts()
+	if err != nil {
 		close(stopChan)
+
+		return "", nil, fmt.Errorf("failed to get forwarded ports: %w", err)
 	}
 
-	return fmt.Sprintf("localhost:%d", localPort), stop, nil
+	var once sync.Once
+
+	stop := func() {
+		once.Do(func() { close(stopChan) })
+	}
+
+	return fmt.Sprintf("localhost:%d", ports[0].Local), stop, nil
 }
--- a/pkg/pod/portforward.go
+++ b/pkg/pod/portforward.go
@@ -1,0 +1,80 @@
+package pod
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	"k8s.io/client-go/tools/portforward"
+	"k8s.io/client-go/transport/spdy"
+	"k8s.io/klog/v2"
+)
+
+const defaultPortForwardReadyTimeout = 60 * time.Second
+
+// PortForward establishes a port-forward to the pod and returns the local address (e.g. "localhost:8443")
+// and a stop function. Callers must invoke the stop function to close the port-forward when done.
+func (builder *Builder) PortForward(localPort, remotePort int) (string, func(), error) {
+	if valid, err := builder.validate(); !valid {
+		return "", nil, err
+	}
+
+	if !builder.Exists() {
+		return "", nil, fmt.Errorf("pod object %s does not exist in namespace %s",
+			builder.Definition.Name, builder.Definition.Namespace)
+	}
+
+	klog.V(100).Infof("Setting up port-forward %d:%d to pod %s in namespace %s",
+		localPort, remotePort, builder.Object.Name, builder.Object.Namespace)
+
+	restConfig := builder.apiClient.Config
+
+	apiURL, err := url.Parse(restConfig.Host)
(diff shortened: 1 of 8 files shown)


pkg/pod/portforward.go
```diff
@@ -0,0 +1,80 @@
...
+
+		return "", nil, fmt.Errorf("port-forward to pod %s did not become ready in %s",
+			builder.Object.Name, defaultPortForwardReadyTimeout)
+	}
+
+	stop := func() {
+		close(stopChan)
+	}
+
+	return fmt.Sprintf("localhost:%d", localPort), stop, nil
```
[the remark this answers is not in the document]
frajamomo: Fixed

pkg/pod/portforward.go
```diff
@@ -0,0 +1,80 @@
...
+	case <-time.After(defaultPortForwardReadyTimeout):
+		close(stopChan)
+
+		return "", nil, fmt.Errorf("port-forward to pod %s did not become ready in %s",
+			builder.Object.Name, defaultPortForwardReadyTimeout)
+	}
+
+	stop := func() {
+		close(stopChan)
+	}
```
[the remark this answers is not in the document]
frajamomo: Fixed

pkg/pod/portforward.go
```diff
@@ -0,0 +1,90 @@
...
+
+	restConfig := builder.apiClient.Config
+
+	apiURL, err := url.Parse(restConfig.Host)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to parse API server URL: %w", err)
+	}
+
+	apiURL.Path = fmt.Sprintf("/api/v1/namespaces/%s/pods/%s/portforward",
+		builder.Object.Namespace, builder.Object.Name)
```
klaskosk: for better symmetry with the pod exec methods, the request builder could be used to get URL:

```go
req := builder.apiClient.CoreV1Interface.RESTClient().
    Post().
    Namespace(builder.Object.Namespace).
    Resource("pods").
    Name(builder.Object.Name).
    SubResource("portforward")
req.URL()  // equivalent to apiURL
```

pkg/pod/portforward.go
```diff
@@ -0,0 +1,90 @@
...
+
+	transport, upgrader, err := spdy.RoundTripperFor(restConfig)
```
klaskosk: maybe there's an opportunity to share some of the roundtripper setup with the pod exec commands? `getExecutorFromRequest` has similar logic, though the function ultimately creates an executor rather than dialer

pkg/pod/portforward.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to create SPDY round-tripper: %w", err)
+	}
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, http.MethodPost, apiURL)
```
klaskosk: optional, but using websockets falling back to SPDY is a more modern approach for exec/portforward. I know for pod exec it tends to be more reliable, but not sure the significance for port fordwarding

```go
spdyDialer := spdy.NewDialer(upgradeRoundTripper, &http.Client{Transport: wrapper}, "POST", req.URL())
wsDialer, err := portforward.NewSPDYOverWebsocketDialer(req.URL(), builder.apiClient.Config)
// ...

dialer := portforward.NewFallbackDialer(wsDialer, spdyDialer, func(err error) bool {
    return httpstream.IsUpgradeFailure(err) || httpstream.IsHTTPSProxyError(err)
}), nil
```

klaskosk (commented):
thanks for including this in eco-goinfra too! I added a couple comments to see if we could share some of the logic from pod exec since it's pretty similar. otherwise looks good

another possible extension would be adding tests to `integration/pod.go`, although port forwarding not as widespread as pod exec so integration tests aren't really a requirement

klaskosk (approved):

PR #1539, opened 2026-07-07
ci: update linter and pin actions by digest

klaskosk:
This prevents the possibility of the floating tag being shifted to a
version with malicious code. Now that all actions are converted,
dependabot will respect the digests.

This commit updates golangci-lint to 2.12.2 as well as fixes the
installation script for the new version. Additional changes are
necessary to appease the updated linters.

Assisted-by: Cursor


klaskosk:
eco-gotests integration is expected to fail

PR #1541, opened 2026-07-08
lca: add IBI installation config helper for preinstall tests

josclark42:
Expose installer-compatible image-based-installation-config.yaml generation through pkg/lca using assisted NetConfig from pkg/schemes, so eco-gotests can avoid direct openshift/assisted-service and openshift/installer imports.

Assisted by: Cursor

## Summary by CodeRabbit

* **New Features**
  * Added support for generating image-based installation configuration files.
  * Included automatic handling for compacting ignition override data and deriving a seed version from an image reference.
  * Wrote the generated configuration to the expected output file with secure permissions.

* **Tests**
  * Added coverage for configuration normalization, seed version extraction, and end-to-end file generation.

### DIFF
--- a/pkg/lca/ibiconfig.go
+++ b/pkg/lca/ibiconfig.go
@@ -1,0 +1,170 @@
+package lca
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	assistedv1 "github.com/rh-ecosystem-edge/eco-goinfra/pkg/schemes/assisted/api/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	// ImageBasedInstallationConfigVersion is the apiVersion for image-based-installation-config.yaml.
+	ImageBasedInstallationConfigVersion = "v1beta1"
+
+	ibiConfigFileName = "image-based-installation-config.yaml"
+)
+
+// ImageDigestSource defines a source repository and optional mirrors for release-image content
+// in image-based-installation-config.yaml (aligned with openshift/installer/pkg/types.ImageDigestSource).
+type ImageDigestSource struct {
+	Source  string   `json:"source"`
+	Mirrors []string `json:"mirrors,omitempty"`
+}
+
+// InstallationConfig is the API for image-based-installation-config.yaml consumed by openshift-install.
+type InstallationConfig struct {
+	metav1.TypeMeta   `json:",inline"`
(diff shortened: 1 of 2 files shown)


pkg/lca/ibiconfig_test.go
```diff
@@ -0,0 +1,93 @@
...
+	raw, err := os.ReadFile(filepath.Join(dir, ibiConfigFileName))
+	require.NoError(t, err)
+
+	content := string(raw)
+	assert.Contains(t, content, "apiVersion: v1beta1")
+	assert.Contains(t, content, "kind: ImageBasedInstallationConfig")
+	assert.Contains(t, content, "seedVersion: 4.16.0")
+	assert.Contains(t, content, "installationDisk: /dev/disk/by-id/wwn-123")
+	assert.Contains(t, content, "interfaces: []")
+}
```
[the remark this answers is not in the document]
josclark42: Addressed in latest commit.

pkg/lca/ibiconfig.go
```diff
@@ -0,0 +1,193 @@
...
+
+// ImageDigestSource defines a source repository and optional mirrors for release-image content
+// in image-based-installation-config.yaml (aligned with openshift/installer/pkg/types.ImageDigestSource).
+type ImageDigestSource struct {
+	Source  string   `json:"source"`
+	Mirrors []string `json:"mirrors,omitempty"`
+}
+
+// InstallationConfig is the API for image-based-installation-config.yaml consumed by openshift-install.
+type InstallationConfig struct {
```
klaskosk, line 32: is this type copied from the installer repo? if so, I think it'd be good to use the existing [sync system](https://github.com/rh-ecosystem-edge/eco-goinfra/tree/main/internal/sync/configs) in eco-goinfra if possible. that'll make it possible to automatically sync the type or at least show more explicit traceability

pkg/lca/ibiconfig.go
```diff
@@ -0,0 +1,193 @@
...
+	InstallationDisk       string                `json:"installationDisk"`
+	NetworkConfig          *assistedv1.NetConfig `json:"networkConfig,omitempty"`
+	PullSecret             string                `json:"pullSecret"`
+	SeedImage              string                `json:"seedImage"`
+	SeedVersion            string                `json:"seedVersion"`
+	SSHKey                 string                `json:"sshKey,omitempty"`
+}
+
+// InstallationConfigInput holds values used to build image-based-installation-config.yaml.
+type InstallationConfigInput struct {
```
klaskosk, line 50: i'm wondering if a better approach here may be leaving the config input in eco-gotests and then only having the types for InstallationConfig in a pkg/schemes package in eco-goinfra?

eco-goinfra does have a few builders which don't correspond directly to CRs, but they still follow the builder structure for setting options. i think taking that approach here would be more effort than it's worth

PR #1548, opened 2026-07-14
secret: stop printing secret data

klaskosk:
This commit redacts the secret data from the WithData and WithStringData methods. Though it may be useful to see this data, secret data should be considered sensitive by default and not printed to the console.

Callers may access the data directly if they wish to store it, rather than printing by default.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive secret values are now redacted from logs, reducing the risk of accidental data exposure while preserving existing behavior.


trewest (approved):

klaskosk:
Thanks Trey!

PR #1573, opened 2026-07-29
assisted: add List functions for Agent

klaskosk:
This commit adds two different list functions for the Agent resource. The first is a generic ListAgents that lists agents based on the controller-runtime client's ListOption. The second is a more specific function which lists the Agents that would appear in the oran-o2ims cluster API.

Assisted-by: Cursor


cdvultur (approved):
/lgtm

PR #1594, opened 2026-08-06
Add CAPI Cluster builder using common builder pattern

frajamomo:
Add typed scheme for cluster.x-k8s.io/v1beta1 Cluster and a builder using the common builder framework. Types are defined locally to avoid pulling in the sigs.k8s.io/cluster-api dependency.

## Summary by CodeRabbit

- **New Features**
  - Added support for creating and retrieving Cluster API cluster resources through a fluent builder.
  - Added configuration for control-plane endpoints, control-plane references, infrastructure references, and paused state.
  - Added validation to prevent invalid configuration values from being applied.

- **Configuration**
  - Added disabled synchronization configuration for Cluster API v1beta1 resources.


pkg/schemes/capi/cluster/v1beta1/groupversion_info.go
```diff
@@ -0,0 +1,25 @@
+// Package v1beta1 contains API Schema definitions for the cluster.x-k8s.io v1beta1 API group.
```
klaskosk, line 1: please add a sync config, even if `sync: false` to track where these types came from

I checked the capi `go.mod` and it looks like you may be able to import them directly if you find a version that has the k8s libs on 0.35 still. or if you could figure out the 0.35 -> 0.36 upgrade. but that's optional, I'm fine if schemes are the easiest approach for now

klaskosk (approved):
lgtm, thanks for adding the sync config! I'll merge once CI finishes

PR #1616, opened 2026-09-09
pkg/reportxml: Add suite-level properties to XML reports

fredericlepied:
- Introduce `Option` and `createOptions` structs for configurable report creation
- Add `WithSuiteProperties` to include caller-supplied properties in XML output
- Adjust `Create` and `setTestSuite` functions to support suite properties
- Implement unit tests to cover new suite-level properties functionality

## Summary by CodeRabbit

- **New Features**
  - Added support for including custom suite-level properties in generated report XML files.
  - Added an option for configuring property name/value pairs when creating reports.
  - Suite properties are emitted in a consistent name-sorted order, while entries without names are omitted.

### DIFF
--- a/pkg/reportxml/reportxml.go
+++ b/pkg/reportxml/reportxml.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/kelseyhightower/envconfig"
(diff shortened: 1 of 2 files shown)


klaskosk (approved):
i like this change, I think we've neglected reportxml a little (we don't even use the output in FT RAN), but this is a positive improvement

